PR #141, opened 2015-01-11
Add tests and implementation for a "quotemark" printer option

weswigham:
@sebmarkbage on the React repo mentioned needing Recast to support a feature like this for them to support configurable quotemarks when compiling JSX transformations in the future. ([mentioned originally on pull request 1709](https://github.com/facebook/react/pull/1709))

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -49,7 +49,12 @@ var defaults = {
 
     // If you want esprima not to throw exceptions when it encounters
     // non-fatal errors, keep this option true.
-    tolerant: true
+    tolerant: true,
+    
+    // If you want to override the quotemarks used in string literals, specify
+    // either "single" or "double" here
+    // Otherwise, the input marks will be preserved
+    quotemark: null,
 }, hasOwn = defaults.hasOwnProperty;
 
 // Copy options and fill in default values.
@@ -73,6 +78,7 @@ exports.normalize = function(options) {
         inputSourceMap: get("inputSourceMap"),
         esprima: get("esprima"),
         range: get("range"),
-        tolerant: get("tolerant")
+        tolerant: get("tolerant"),
+        quotemark: get("quotemark"),
     };
 };
--- a/lib/options.js
+++ b/lib/options.js
@@ -51,10 +51,11 @@ var defaults = {
     // non-fatal errors, keep this option true.
     tolerant: true,
     
-    // If you want to override the quotemarks used in string literals, specify
-    // either "single" or "double" here
+    // If you want to override the quotes used in string literals, specify
+    // either "single", "double", or "auto" here ("auto" will select the one 
+    // which results in the shorter literal)
     // Otherwise, the input marks will be preserved
-    quotemark: null,
+    quote: null,
 }, hasOwn = defaults.hasOwnProperty;
 
 // Copy options and fill in default values.
@@ -79,6 +80,6 @@ exports.normalize = function(options) {
         esprima: get("esprima"),
         range: get("range"),
         tolerant: get("tolerant"),
-        quotemark: get("quotemark"),
+        quote: get("quote"),
     };
 };
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -126,13 +126,6 @@ function findReprints(newPath, reprints, options) {
 function findAnyReprints(newPath, oldPath, reprints, options) {
     var newNode = newPath.value;
     var oldNode = oldPath.value;
-        
-    if (options && options.quotemark &&
-          newPath.parentPath && 
-          newPath.parentPath.value &&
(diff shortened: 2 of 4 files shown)


abuiles:
@weswigham thanks! @benjamn it would be great to see this is in, we are using recast to generate code in https://github.com/ember-cli/ember-cli and people keep complaining that sometimes we generate with double quote even though single quote is our recommendation.

lib/options.js
```diff
@@ -49,7 +49,12 @@ var defaults = {
 
     // If you want esprima not to throw exceptions when it encounters
     // non-fatal errors, keep this option true.
-    tolerant: true
+    tolerant: true,
+    
+    // If you want to override the quotemarks used in string literals, specify
+    // either "single" or "double" here
+    // Otherwise, the input marks will be preserved
+    quotemark: null,
```
benjamn: Can you change this so that `options.quotemark` is either `null`, `'"'` (rather than `"double"`), or `"'"` (rather than `"single"`)?

lib/patcher.js
```diff
@@ -123,23 +123,30 @@ function findReprints(newPath, reprints) {
...
     var newNode = newPath.value;
     var oldNode = oldPath.value;
-
+        
+    if (options && options.quotemark &&
+          newPath.parentPath && 
+          newPath.parentPath.value &&
+          newPath.parentPath.value.type === "Literal" && 
+          isString.check(newPath.parentPath.value.value))
+        return false; //mark literal for reprint
```
benjamn: This is a bit aggressive: if the code wasn't modified by the traversal, then its quoting style should be left intact, I think. As long as you've implemented the printing code to use the preferred style for new nodes and nodes that need to be reprinted, I think you'll get the effect you want. You can always set `node.original = null` during the traversal to trigger reprinting for that node.

lib/options.js
weswigham: I was using `"double"` and `"single"` to be inline with the jshint/jslint options, but `'"'` and `'"'` are both doable.

abuiles: Maybe we could aim to some parity with escodegen? they use the key "quotes" instead of "quotemark" and "single",  "double" or "auto" as options. see https://github.com/estools/escodegen/wiki/API#optionformatquotes

weswigham: I like mirroring that - though I think auto and null/undefined should be the same?

weswigham:
I removed the over-aggressive literal re-emissions from the patcher, and mirrored the escodegen quote transform api (since, on reflection, selecting the shortest literal makes sense for something like a minifier).

abuiles:
:+1:

kiwiupover:
:+1: Thanks

lib/patcher.js
```diff
@@ -340,7 +340,7 @@ function findChildReprints(newPath, oldPath, reprints) {
         if (k === "loc")
             continue;
 
-        if (!findAnyReprints(newPath.get(k), oldPath.get(k), reprints))
+        if (!findAnyReprints(newPath.get(k), oldPath.get(k), reprints, options))
```
benjamn: I might be crazy, but I can't tell where `options` is used in this file. It seems like it's just being passed through the various `find*` functions but never examined. Am I missing something obvious?

lib/patcher.js
```diff
@@ -123,23 +123,23 @@ function findReprints(newPath, reprints) {
     return canReprint;
 }
 
-function findAnyReprints(newPath, oldPath, reprints) {
+function findAnyReprints(newPath, oldPath, reprints, options) {
     var newNode = newPath.value;
     var oldNode = oldPath.value;
-
+
```
benjamn: Nit: trailing whitespace.

lib/printer.js
```diff
@@ -1283,10 +1283,26 @@ function endsWithBrace(lines) {
...
+        return m === '"' ? '\'' : '"';
+    });
+}
+
+function nodeStr(n, options) {
     namedTypes.Literal.assert(n);
     isString.assert(n.value);
-    return JSON.stringify(n.value);
+    switch (options.quote) {
+        case "auto":
```
benjamn: Nit: line up `case`s vertically with `switch` (de-indent them please).

benjamn:
Made some comments; sorry for the delay!

lib/patcher.js
weswigham: You're right! After I removed the overzealous reprinting it wasn't used anymore. Let me clean this up.

weswigham:
Removed all the stale changes and changed the whitespace of the switch to your preference.

abuiles:
@weswigham @benjamn thanks! https://github.com/ember-cli/ember-cli/pull/3182 :+1:

Issue #139, opened 2014-12-31
Experiment: stop using NodePath in favor of a simple […, name, value] stack


benjamn:
Nothing in Recast really depends on the `NodePath` abstraction, and printing is a read-only operation over the AST, anyway, so I think we can get away with something cheaper in https://github.com/benjamn/recast/blob/master/lib/printer.js and https://github.com/benjamn/recast/blob/master/lib/patcher.js.

The logic behind `NodePath.prototype.needsParens` will need to be rewritten to operate on a stack of name/value elements, but that should be straightforward.

cc @eventualbuddha @stefanpenner @drslump

briandipalma:
I think the printing has been affected by this change. Comments seem to be printed twice now.

```javascript
/**
 * This class was in an IIFE and returned an instance of itself.
 */
function SimpleClass() {
}
```

Take this as input and add a statement to the end of it so it should look like

```javascript
/**
 * This class was in an IIFE and returned an instance of itself.
 */
function SimpleClass() {
}
module.exports = SimpleClass;
```

And the comment is printed twice, I'm trying to debug this now but even with this simple sample it's not clear where the bug is. Is `FastPath` `stack` meant to contain AST nodes?

briandipalma:
Everything is fine until the patcher decides to concat the comment to the top of the lines array.

benjamn:
Can you submit a quick PR with a failing test case, or just paste the test case here?

briandipalma:
```javascript
            if (comparePos(sliceFrom, rep.start) > 0) {
                // Ignore nested replacement ranges.
            } else {
//                pushSlice(sliceFrom, rep.start);
                toConcat.push(rep.lines);
                sliceFrom = rep.end;
            }
```

Commenting out the `pushSlice` seems to stop it. It's quite late here so I'll have to head to bed but I'll try and do that tomorrow.

briandipalma:
I have created a PR with a failing test https://github.com/benjamn/recast/pull/149

I'd prefer to just fix the issue but it's not clear what exactly is going wrong. Maybe the indexes are off in the `Patcher`.

benjamn:
@briandipalma I believe that issue is fixed with https://github.com/benjamn/recast/tree/v0.9.18, published just now to NPM.

briandipalma:
@benjamn Thanks, fixed the issue!

benjamn:
`FastPath` has been working quite nicely for several months now!

stefanpenner:
[...]

PR #144, opened 2015-01-17
Remove types.NodePath dependency.

benjamn:
This implements the experiment proposed in #139.


benjamn:
These changes result in a noticeable speedup (~30%) for `recast.print`, and that's without much profiling and optimization.

PR #145, opened 2015-01-23
Fix "else if" check in example

milkbikis:


milkbikis:
I'm not quite sure why this is failing the syntax test. FWIW, it fails locally when I reset to earlier commits as well. Let me know if I need to look into the test failure.

benjamn:
Ah yes, that's not your fault; that's just because https://github.com/benjamn/ast-types was upgraded recently, without a proper update to Recast.

milkbikis:
Thanks!

PR #146, opened 2015-01-29
fixes #128 parse inline sourcemap when parsing

### LINKED ISSUE #128: consume sourcemap directive when parsing source code
according to our conversation yesterday, if the sourcemap is base64 embeded in the source code, we should do this:
- extract it before calling esprima (for perf reasons)
- digest the sourcemap as the input sourcemap

caridy:
* extract inline base64 sourcemap from the source before parsing code with esprima
* basic regex to match `sourceMappingURL` from the end of the source (we probably should make this more restrictive)
* you can overrule the input sourcemap using `options.inputSourceMap` when printing.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -9,11 +9,13 @@ var Patcher = require("./patcher").Patcher;
 var normalizeOptions = require("./options").normalize;
 var fromString = require("./lines").fromString;
 var attachComments = require("./comments").attach;
+var extractInlineSourceMaps = require('./util').extractInlineSourceMaps;
 
 exports.parse = function parse(source, options) {
     options = normalizeOptions(options);
 
-    var lines = fromString(source, options);
+    var map = extractInlineSourceMaps(source);
+    var lines = fromString(map.source, options);
 
     var sourceWithoutTabs = lines.toString({
         tabWidth: options.tabWidth,
@@ -48,6 +50,14 @@ exports.parse = function parse(source, options) {
 
     // Attach comments to the copy rather than the original.
     attachComments(comments, copy.program, lines);
+
+    // Attach inline sourcemap metadata when possible
+    Object.defineProperty(copy, 'inputSourceMap', {
+        enumerable: false,
+        configurable: false,
+        writable: false,
+        value: map.inputSourceMap
+    });
 
     return copy;
 };
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -109,7 +109,7 @@ function Printer(originalOptions) {
         return new PrintResult(
             lines.toString(options),
             util.composeSourceMaps(
-                options.inputSourceMap,
+                options.inputSourceMap || ast.inputSourceMap,
                 lines.getSourceMap(
                     options.sourceMapName,
                     options.sourceRoot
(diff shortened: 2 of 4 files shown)


lib/parser.js
```diff
@@ -49,6 +51,14 @@ exports.parse = function parse(source, options) {
     // Attach comments to the copy rather than the original.
     attachComments(comments, copy.program, lines);
 
+    // Attach inline sourcemap metadata when possible
+    Object.defineProperty(copy, 'inputSourceMap', {
```
caridy: we could just a different way to annotate the ast, for now, I'm using this non-enumerable property.

lib/printer.js
```diff
@@ -109,7 +109,7 @@ function Printer(originalOptions) {
         return new PrintResult(
             lines.toString(options),
             util.composeSourceMaps(
-                options.inputSourceMap,
+                options.inputSourceMap || ast.inputSourceMap,
```
caridy: preserving the current behavior where `options.inputSourceMap` has priority.

lib/util.js
```diff
@@ -75,3 +75,13 @@ exports.composeSourceMaps = function(formerMap, latterMap) {
 
     return smg.toJSON();
 };
+
+exports.extractInlineSourceMaps = function (source) {
+    var re = /\s*\/\/(?:@|#) sourceMappingURL=data:application\/json;base64,(\S*)$/m,
```
caridy: this can be improved

benjamn:
Awesome, I look forward to reviewing this!

test/mapping.js
```diff
@@ -169,6 +169,95 @@ describe("source maps", function() {
...
+
+        var ast = parse(code, {
+            sourceFileName: "original.js"
+        });
+
+        var useStrictResult = new Printer({
+            sourceMapName: "useStrict.map.json"
+        }).print(addUseStrict(ast));
+
+        var useStrictResultCodeAndMap = [
```
caridy: same as previous test, but passing the map as inline in base64.

caridy:
@benjamn any update on this?

benjamn:
Hey @caridy, sorry it's taking so long for me to get to this. I'm tempted to go ahead and merge this, but what I'd really prefer is to associate the input source map with the original `Lines` object, so that when you call `getSourceMap` on the final printed `Lines` object, the source map you get is composed with the input map.

If I end up changing the API to work that way in the near future, will it be a problem for you? Can you avoid relying on the `ast.inputSourceMap` property in your own code? In other words, I hope the storage of the input source map can remain an implementation detail of how `recast.parse` and `recast.print` communicate with each other.

caridy:
@benjamn not at all, do what you think is best, we are not relying on this features at any level today.

demoneaux:
Any updates?

lib/util.js
flotwig: i know this is a 5-year old PR, but i don't think this will work with URLs with a charset specified, like `data:application/json;charset=utf-8;base64,...`

Qix-:
Hi, just another poke. This is sorely missed functionality...

PR #149, opened 2015-02-07
Add test for printer double comments bug.

briandipalma:
Adding a test for double comments bug. The comments are being double in the `Patcher` class but I'm not sure what the root issue is.

### DIFF
--- a/test/printer.js
+++ b/test/printer.js
@@ -685,8 +685,8 @@ describe("printer", function() {
             " * This class was in an IIFE and returned an instance of itself." +
             " */" +
             "function SimpleClass() {" +
-            "};" +
+            "}\n" +
             "module.exports = SimpleClass;"
-        ].join("\n"));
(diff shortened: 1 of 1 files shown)


benjamn:
Perfect, thanks! I'll take a look at this today.

briandipalma:
Hi, is there some way to shortcircuit the performance improvements? I've tried setting `original` to undefined on the program nodes and it still seems to be printing the comments where they used as opposed to where the AST says the should be. Performance isn't a concern for me, the AST is correct when I look at it in the debugger so I was curious if there was some to make recast thing this is a fresh AST with no history?

benjamn:
Try `recast.prettyPrint(ast).code`?

briandipalma:
Umm, `prettyPrint` seems to strip away all the comments. I guess I can manually move the comments around for the moment.

Issue #151, opened 2015-02-12
Why are comment nodes not `configurable`?


briandipalma:
As part of the code converter I'm building we'd like our comments to appear in different places in the finished AST. For example:

```javascript
// Some doc.
my.namespace.Class = (function() {

function Class() {}

return Class;
});
```

Should be converted to

```javascript
// Some doc.
function Class() {}

module.exports = return Class;
```

This is easy enough in theory, I take the `Program` comment and attach it to the constructor function declaration node and remove it from the `Program` node. Sadly comments are not `configurable` so I can't define them to be `undefined` so the Program comment is printed twice.

I was wondering if there was a reason for this choice or was it just a case of using the default behaviour of `defineProperty`.

briandipalma:
PR if this can be changed https://github.com/benjamn/recast/pull/152

benjamn:
Quick update: the `.comment` property is now a simple (enumerable, writable, configurable) object property, created without using `Object.defineProperty` at all: https://github.com/benjamn/recast/commit/b35be9aa74ad4c51234054bf4d1cbc904f9a9534

briandipalma:
OK, is this new release meant to address the double comment bug? Is there are new feature (reved the minor number I noticed)? Thanks!

briandipalma:
Issues fixed, excellent thank you very much!

PR #152, opened 2015-02-12
Make comments node configurable. Issue #151

briandipalma:


benjamn:
Let's make it `writable: true` too, so that you don't have to use `Object.defineProperty` to change it.

jugglinmike:
I'm having trouble modifying comments, either through re-setting the property wholesale (with this path applied) or through modifying the properties of the object created by Recast. In either case, `recast.print` retains the comments in original form when I pass the ast to `recast.print`. I can't find an example of (or test for) comment modification in this project, so I'm not sure if I'm missing something or if this is not a supported feature. Can anyone advise?

benjamn:
@jugglinmike That's a bug that I've just recently become aware of, and it's my top priority now: https://github.com/benjamn/recast/issues/153. Thanks for speaking up about it.

jugglinmike:
@benjamn Thanks for being on top of it :)

briandipalma:
Sorry, I made this PR before starting work so I've not been able to respond but I'm also experiencing the issues @jugglinmike has mentioned. I was planning to write some tests to show the failure before raising the bug :)

Looks like you are already on it anyway!

PR #161, opened 2015-03-02
fix arrow function print logic to better account for rest params

blazarus:
If an arrow function has one normal parameter and one rest parameter, recast will leave out the rest param when printing the arrow function. I've fixed the logic in printer.js and modified an existing unit test to exercise this code path.

**Note** this fix depends on this fix in ast-types: https://github.com/benjamn/ast-types/commit/39ac68d61b1a6c211245c87141c335ee7a2923a9
(apologies if there is a better way to express this dependency, or if I should be waiting to issue this pull request until the corresponding fix in ast-types is integrated)

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -313,6 +313,9 @@ FPp.needsParens = function(assumeExpressionContext) {
             return false;
         }
 
+    case "ArrowFunctionExpression":
+        return isBinary(parent);
+
     default:
         if (parent.type === "NewExpression" &&
             name === "callee" &&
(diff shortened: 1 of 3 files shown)


benjamn:
Thanks for this! I've published `ast-types` v0.7.0 with that change, so you should be able to depend on the NPM version now.

blazarus:
@benjamn awesome, thanks!

benjamn:
Ping? No big hurry, but it looks like this needs a rebase, and you might as well switch back to `"ast-types": "~0.7.0"` in `package.json` :)

blazarus:
Ok, rebased in the latest changes - sorry for the delay. Are you okay with two commits or would you prefer I squashed them?

benjamn:
Two commits is fine with me. Thanks!

PR #168, opened 2015-03-05
Allow ExpressionStatement and BlockStatement to not add semicolons with the pretty printer

justafish:
I've only implemented this for ExpressionStatement and BlockStatement so far, but it'd be nice to have trailing semicolon addition as optional.


justafish:
I think actually what I'd really like to be able to do is override the printer for specific types, e.g. for ConditionalExpression I would prefer to not have parenthesis, so I could just supersede that particular output with my own.

Let me know what you think

eventualbuddha:
Closing as stale.

PR #169, opened 2015-03-05
handle parentheses around object literals in arrow functions correctly

blazarus:
In es6 arrow functions like this `(x, y) => x + y` the return statement is implicit. However, if the thing you're returning is an object literal, you need to surround it with parentheses (since curly braces are otherwise used to denote a function body).

Recast should add parentheses correctly in this situation.

Note: I wasn't sure where to add a unit test for this, but would be happy to do so if pointed in the right direction.

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -316,6 +316,12 @@ FPp.needsParens = function(assumeExpressionContext) {
     case "ArrowFunctionExpression":
         return isBinary(parent);
 
+    case "ObjectExpression":
+        if (parent.type === "ArrowFunctionExpression" &&
+            name === "body") {
+            return true;
+        }
+
     default:
         if (parent.type === "NewExpression" &&
             name === "callee" &&
--- a/test/parens.js
+++ b/test/parens.js
@@ -116,6 +116,9 @@ describe("parens", function() {
         check("(function(){}).apply(this, arguments)");
         check("function f() { (function(){}).call(this) }");
         check("while (true) { (function(){}).call(this) }");
+        check("() => ({a:1,b:2})");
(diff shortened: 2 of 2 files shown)


blazarus:
sorry, looks like I have a failing unit test. Closing this until I fix that.

lib/fast-path.js
```diff
@@ -316,6 +316,9 @@ FPp.needsParens = function(assumeExpressionContext) {
     case "ArrowFunctionExpression":
         return isBinary(parent);
 
+    case "ObjectExpression":
+        return parent.type === "ArrowFunctionExpression";
```
benjamn: Can you add `&& name === "body"` to the returned expression here? There's a small possibility that the object expression could be a default initializer expression, as in `(x, y={z:1}) => x + y.z`, and I don't think we'd want to force that `{z:1}` to be wrapped in parentheses.

benjamn:
Feel free to keep it open even if the tests are failing! I won't merge it until it's ready :)

benjamn:
An easy way to test this is to build an `ArrowFunctionExpression` using `recast.types.builders.*`, then call `recast.prettyPrint(node, { tabWidth: 2 }).code` and string-compare that to the expected output. It's a little fragile, because printing details can change from time to time, but as long as the expression is small, I think it's a useful test.

lib/fast-path.js
blazarus: good call! Will do.

blazarus:
Thanks. I ended up adding checks in test/parens.js, which seems to be doing something similar to your suggestion above. (I also added a check for my prior commit https://github.com/benjamn/recast/commit/d669a9eddad69304961a7f4d8a85159285b5d9fd)

PR #173, opened 2015-03-22
Add ability to print Flow/TypeScript type annotations.

cpojer:
This adds printing for all flow/typescript syntax. I'll outline a few things in the comments. Note that this breaks 3 tests. It might be related to the `needsParens` code, but I'm unsure what to do to make the printing work. It seems that it prints `(function()() {})` now. @benjamn do you have an idea about this?

cc @jeffmo @gabelevi

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -208,7 +208,10 @@ function genericPrintNoParens(path, options, print) {
         return fromString(".").join(n.body);
 
     case "Identifier":
-        return fromString(n.name, options);
+        return concat([
+            fromString(n.name, options),
+            path.call(print, "typeAnnotation")
+        ]);
 
     case "SpreadElement":
     case "SpreadElementPattern":
@@ -229,13 +232,20 @@ function genericPrintNoParens(path, options, print) {
         if (n.generator)
             parts.push("*");
 
-        if (n.id)
-            parts.push(" ", path.call(print, "id"));
+        if (n.id) {
+            parts.push(
+                " ",
+                path.call(print, "id"),
+                path.call(print, "typeParameters")
+            );
+        }
 
         parts.push(
             "(",
             printFunctionParams(path, options, print),
-            ") ",
+            ")",
+            path.call(print, "returnType"),
+            " ",
             path.call(print, "body")
         );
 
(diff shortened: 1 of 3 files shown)


lib/printer.js
```diff
@@ -229,13 +232,20 @@ function genericPrintNoParens(path, options, print) {
         if (n.generator)
             parts.push("*");
 
-        if (n.id)
-            parts.push(" ", path.call(print, "id"));
+        if (n.id) {
+            parts.push(
+                " ",
+                path.call(print, "id"),
+                path.call(print, "typeParameters")
```
cpojer: removing this line and `returnType` on line #247 makes the failing tests pass.

lib/printer.js
```diff
@@ -924,17 +955,38 @@ function genericPrintNoParens(path, options, print) {
         return concat(parts);
 
     case "ClassProperty":
-        return concat([path.call(print, "id"), ";"]);
```
cpojer: this was actually a bug – the field is called `key`, not `id`.

lib/printer.js
```diff
@@ -954,6 +1006,9 @@ function genericPrintNoParens(path, options, print) {
     case "Specifier":
     case "NamedSpecifier":
     case "Comment": // Supertype of Block and Line.
+    case "MemberTypeAnnotation": // Flow
+    case "TupleTypeAnnotation": // Flow
+    case "Type": // Flow
```
cpojer: I couldn't find these three types anywhere. I assume "Type" is just the supertype in ast-types; `TupleTypeAnnotation` is implicit (?) and I don't know what `MemberTypeAnnotation` is – it is not even in esprima, so I should probably just get rid of it here.

lib/printer.js
```diff
@@ -962,10 +1017,227 @@ function genericPrintNoParens(path, options, print) {
...
+    case "DeclareVariable":
+        return concat([
+            fromString("declare var ", options),
+            path.call(print, "id")
+        ]);
+
+    case "FunctionTypeAnnotation":
+        // yay magic. FunctionTypeAnnotation is ambigous:
+        // declare function(a: B): void; OR
+        // var A: (a: B) => void;
```
cpojer: This is the only super-clowntown part of this implementation. It turns out that `FunctionTypeAnnotation` is ambiguous and renders differently based on the context around it.

cpojer:
The only thing I'm unsure about is the formatting, but that shouldn't block this diff. What is the smart way to make it optional for users to decide whether they want `var a: number` or `var a:number`?

The one decision I made in this code is that in an `ObjectTypeAnnotation` the last annotation will receive a semicolon, except when there is only a single annotation which is forced on a single line without a semicolon. This matches what we've been doing at fb (for example: `var obj: {[name: string]: string};` )

lib/printer.js
gabelevi: Member type is deprecated. Type is abstract. Tuple type is

    var x:[number, string] = [123, "hi"];

cpojer:
Alright, added the tuple typecheck and removed the literal check for `nodeStr` so the changes to ast-types are not required.

benjamn:
Rebase? Heads up: I pushed a commit earlier that changed the type of the first parameter to `nodeStr`.

benjamn:
It looks like `class A<T:number> {}` isn't something npmjs.org/package/esprima-fb knows how to parse?

cpojer:
Updated esprima-fb, now it should work just fine.

benjamn:
If you rebase against bb4b4d4fade73d3907b30bcfa5f4467e1f00814a (currently `origin/master`) then I think your tests will pass.

cpojer:
it works! Awesome. Thanks @benjamn

benjamn:
I guess I've just avoided calling `print` on undefined children so far, or I would have run into that bug already.

lib/printer.js
mitermayer: @cpojer, I may be missing something but is there a particular reason why 'TupleTypeAnnotation' is meant to cause an unprintable error? 

I have been hitting this error:

```javascript
Error: unprintable type: "TupleTypeAnnotation"
    at genericPrintNoParens 
```

cpojer: No idea, the last time I touched this was more than two years ago.

mitermayer: @cpojer  thanks for your reply, created an issue here to track it (https://github.com/benjamn/recast/issues/422)

PR #179, opened 2015-05-23
Add parenthesis to arrow function params when using SpreadElementPattern

cpojer:
Before: `...rest => rest`
After: `(...rest) => rest`

I circumvented this by doing

```js
b.arrowFunctionExpression(…);
b.rest = j.identifier('rest');
```

but I prefer to use the SpreadElementPattern here because it means I can just append it to the params list:

```js
j.arrowFunctionExpression(
  (fn.params || []).concat(
    j.spreadElementPattern(j.identifier('rest'))
  ),
  fn.body,
  fn.generator
);
```

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -257,7 +257,11 @@ function genericPrintNoParens(path, options, print) {
         if (n.async)
             parts.push("async ");
 
-        if (n.params.length === 1 && !n.rest) {
+        if (
+            n.params.length === 1 &&
+            !n.rest &&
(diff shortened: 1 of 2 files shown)


cpojer:
Actually this might not make sense because of how esprima is using `node.rest` but I figured this is nice for completeness. Does any parser out there parse rest params this way?

benjamn:
Did you mean `... && !n.rest && ...`?

cpojer:
yes, absolutely! Added a test for this case too.

PR #183, opened 2015-05-28
Print trailing commas when the `trailingComma` option is specified.

cpojer:
This adds support to print trailing commas, which is useful for code bases that adopted this style. They are only added if the corresponding list is split over multiple lines.

I had to add printing logic similar from function arguments to array expressions which, when `wrapColumn` is specified, tries to fit everything within the limit.

Are there any places that I missed where we should add trailing commas?

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -486,7 +486,6 @@ function genericPrintNoParens(path, options, print) {
         fields.forEach(function(field) {
             path.map(function(childPath) {
                 var i = childPath.getName();
-                var prop = childPath.getValue();
                 var lines = print(childPath);
 
                 if (!oneLine) {
@@ -507,6 +506,8 @@ function genericPrintNoParens(path, options, print) {
                     parts.push(separator + (multiLine ? "\n\n" : "\n"));
                     allowBreak = !multiLine;
                 } else if (len !== 1 && isTypeAnnotation) {
+                    parts.push(separator);
+                } else if (options.trailingComma) {
                     parts.push(separator);
                 }
             }, field);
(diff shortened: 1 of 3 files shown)


lib/printer.js
```diff
@@ -556,11 +561,18 @@ function genericPrintNoParens(path, options, print) {
                 // both (all) of the holes.
                 parts.push(",");
             } else {
-                if (i > 0)
-                    parts.push(" ");
-                parts.push(print(elemPath));
-                if (i < len - 1)
+                var lines = print(elemPath);
```
cpojer: @benjamn this now prints items in an array twice – is there a way to reuse the printed data from above? If not, I think it is worth it to iterate over the elements twice: once to check if there are holes in the array and once for printing. It would be a little bit more code though.

cpojer:
@benjamn ping? :)

lib/printer.js
benjamn: Oh, yeah, reusing the printed elements (`Lines` objects) from above should be feasible.

cpojer:
Alright, sorry @benjamn that this took so long – I'm now reusing the printed array values.

lib/printer.js
```diff
@@ -1486,6 +1502,9 @@ function printFunctionParams(path, options, print) {
     if (joined.length > 1 ||
         joined.getLineLength(1) > options.wrapColumn) {
         joined = fromString(",\n").join(printed);
+        if (options.trailingComma && !fun.rest) {
```
benjamn: Does this mean there won't be a trailing comma after a `...rest` parameter?

cpojer: yeah, seems like `foo(a, ...b,) {}` is not legal atm.

benjamn: Ah, yes, I remember a discussion at TC39 about this: you're never going to add another parameter after a `...rest`, so the argument from minimizing diff lines does not apply.

test/printer.js
```diff
@@ -797,6 +797,85 @@ describe("printer", function() {
...
+            "  1,",
+            "  2,",
+            ");"
+        ].join("\n");
+
+        var ast = parse(code);
+
+        var printer = new Printer({
+            tabWidth: 2,
+            wrapColumn: 1,
```
benjamn: Now I'm glad I added `options.wrapColumn` :)

PR #186, opened 2015-06-15
Attempt to guess whether or not to use tabs

### LINKED ISSUE #185: Infer options.useTabs when not specified
It would be awesome if recast inferred `options.useTabs` from the file contents, just like it does with `options.tabWidth` right now.

Right now, in a tab-indented file with `useTabs` not set to true, newly injected statements will be space-indented (using, I believe, a 4-space tab width by default). It would be great if `useTabs` were inferred as being true for newly injected statements in that case.

platinumazure:
Attempting to fix #185. Unfortunately that's not happening just yet. New statements start with no indentation, then get spaces added despite the file having been guessed as `{ useTabs: true }`.

Could use some help debugging this, to be honest. I definitely feel like I'm on the right track, but I'm just missing something in the recursive (and hard to trace) print flow. But if anyone has a better approach, I'm open to suggestions.

### DIFF
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -483,7 +483,7 @@ Lp.guessUseTabs = function() {
 
     var result = tabIndentCount > noTabIndentCount;
 
-    return secret.cachedUseTabs = result
+    return secret.cachedUseTabs = result;
 };
 
 Lp.isOnlyWhitespace = function() {
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -476,7 +476,7 @@ Lp.guessUseTabs = function() {
         // Intentionally not counting mixed spaces/tabs
         if (hasTab && !hasSpace) {
             ++tabIndentCount;
-        } else if (hasSpace && !hasTab) {
+        } else {
             ++noTabIndentCount;
         }
     }
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -426,6 +426,66 @@ Lp.guessTabWidth = function() {
     return secret.cachedTabWidth = result;
 };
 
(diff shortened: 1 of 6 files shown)


platinumazure:
I'm thinking I want to look at [`printStatementSequence`][1], does that sound about right?

  [1]: https://github.com/benjamn/recast/blob/master/lib/printer.js#L1282

benjamn:
Thanks for digging into this problem, and especially for understanding so much of the `Lines` code.

This problem appears to be harder than I thought, since guessing the tab width (i.e., the number of spaces used for one unit of indentation) is tricky when there are any tabs in the file. One way to simplify it might be to require `options.tabWidth` to be provided explicitly in order to guess `options.useTabs`. Does that seem reasonable?

benjamn:
What I mean is: guessing both `options.useTabs` and `options.tabWidth` is a deceptively difficult problem, which might be why text editors like TextMate and Sublime require the user to specify the tab width manually.

platinumazure:
> 

What I mean is: guessing both options.useTabs and options.tabWidth is a deceptively difficult problem, which might be why text editors like TextMate and Sublime require the user to specify the tab width manually.

You're probably not wrong about that... That said, we do tend to know/infer `options.tabWidth`, so maybe `Lines.prototype.indent` could take in an extra argument for tab width and (if that argument is specified) convert to tabs accordingly?

Or, something else to think about, could the `Lines` instances themselves store a tabWidth and/or useTabs value and just retain that? Then calls to `indent`/`indentTail`/similar wouldn't need to know if tabs are on or off.

Issue #190, opened 2015-07-02
Trailing block comments are getting semicolons


zertosh:
```js
var recast = require('recast');

var ast = recast.parse([
  'module.exports = {};',
  '/**',
  ' * trailing comment',
  ' */'
].join('\n'));

recast.visit(ast, {
  visitNode: function(path) {
    this.traverse(path);
    path.value.original = null;
  }
});

console.log(recast.print(ast).code);
```

Actual:
```
module.exports = {};
/**
 * trailing comment
 */;
```

Expected:
```
module.exports = {};
/**
 * trailing comment
 */
```

benjamn:
This should be fixable. Want to submit a PR with a failing test case?

zertosh:
Thanks! see https://github.com/benjamn/recast/pull/192

PR #192, opened 2015-07-02
Add failing test for trailing body comments

zertosh:
see https://github.com/benjamn/recast/issues/190

cc: @benjamn


zertosh:
I didn't add it "TrailingComments" because it would fail the `astNodesAreEquivalent` test (the block comment turns into an `EmptyStatement` with the semicolon), and that's a really cryptic error.

zertosh:
Thanks for the super quick fixes @benjamn!

PR #193, opened 2015-07-02
Use mocha@~2.2.5

zertosh:
Includes a fix for seeing inline diffs (https://github.com/mochajs/mocha/pull/1655).

rbass


zertosh:
cc: @benjamn

PR #196, opened 2015-07-09
Printing support for property initializers.

### LINKED ISSUE #188: Support for ES7 Property Initializers
Was trying to use `recast` on a react project and ran into an issue with it not being able to parse `static` property initializers. For example:

```
class App extends Component {
  static defaultProps = { /* ... */ };
}
```

Which results in an unexpected token error. Since these are ES7+ I understand why I shouldn't expect this to work, but is there any way for `recast` to parse these initializers and, if not, is there any way that I can add them to `ast-types` to allow me to visit these properties when going through the AST? Thanks!

cpojer:
This adds printing support for property initializers in "ES7".

Depends on https://github.com/benjamn/ast-types/pull/117 and https://github.com/benjamn/recast/pull/179 – I can remove the other recast diff though, just accidentally stacked it on top.
Fixes #188

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1105,11 +1105,19 @@ function genericPrintNoParens(path, options, print) {
         return concat(parts);
 
     case "ClassProperty":
-        return concat([
-            path.call(print, "key"),
-            path.call(print, "typeAnnotation"),
-            ";"
-        ]);
+        var parts = [];
+        if (n.static)
+            parts.push("static ");
+
+        parts.push(path.call(print, "key"));
+        if (n.typeAnnotation)
+            parts.push(path.call(print, "typeAnnotation"));
+
+        if (n.value)
+            parts.push(" = ", path.call(print, "value"));
+
+        parts.push(";");
+        return concat(parts);
 
     case "ClassDeclaration":
     case "ClassExpression":
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -279,7 +279,11 @@ function genericPrintNoParens(path, options, print) {
         if (n.async)
             parts.push("async ");
 
-        if (n.params.length === 1 && !n.rest) {
+        if (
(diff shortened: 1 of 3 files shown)


lib/printer.js
```diff
@@ -1101,11 +1105,19 @@ function genericPrintNoParens(path, options, print) {
...
 
     case "ClassProperty":
-        return concat([
-            path.call(print, "key"),
-            path.call(print, "typeAnnotation"),
-            ";"
-        ]);
+        var parts = [];
+        if (n.static)
+            parts.push("static ");
```
benjamn: I've started listening to my orthodontist and putting braces on everything, though you can be forgiven for following the existing style in this file :)

lib/printer.js
```diff
@@ -279,7 +279,11 @@ function genericPrintNoParens(path, options, print) {
         if (n.async)
             parts.push("async ");
 
-        if (n.params.length === 1 && !n.rest) {
+        if (
+            n.params.length === 1 &&
+            !n.rest &&
+            n.params[0].type !== 'SpreadElementPattern'
```
benjamn: This commit looks reasonable to me, too!

lib/printer.js
cpojer: nice! I'd reserve adding braces for a separate pull request though, just following your conventions. How do you feel about spaces instead of tabs? :P

benjamn: NO TABS EVER

cpojer:
Seems like the tests are failing because ast-types is not updated. I guess I'll have to wait for you to merge the change to ast-types, publish it and then go back to this diff and update package.json?

benjamn:
Want to rebase and see if the tests pass now?

cpojer:
works!

cpojer:
wohoo!

PR #198, opened 2015-07-11
Fix import type/typeof and fix minor type printing issues.

cpojer:
In https://github.com/facebook/esprima/commit/01e5f03ae650a228507504236ceb4ed06372e841 we updated how we use `import type` and added support for `import typeof`. We removed `isKind` and are instead relying on `importKind` which can be any of `value`, `type` or `typeof`. I made it so that it prints  `importKind` directly when it is not `value`.

I noticed also that there were minor bugs in the support for type printing. I discovered this because the ast in both cases was correct but the printed source was not. I changed the test to also directly compare the printed source, which required me to add newlines/spaces to the input source. That seemed better to me compared to using regex to remove newlines and spaces from the output. I will highlight some other fixes inline.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -474,6 +474,10 @@ function genericPrintNoParens(path, options, print) {
     case "ImportDeclaration":
         var parts = ["import "];
 
+        if (n.importKind && n.importKind !== "value") {
+            parts.push(n.importKind + " ");
+        }
+
         if (n.specifiers &&
             n.specifiers.length > 0) {
 
@@ -579,9 +583,9 @@ function genericPrintNoParens(path, options, print) {
         var oneLine = (isTypeAnnotation && len === 1) || len === 0;
         var parts = [oneLine ? "{" : "{\n"];
 
+        var i = 0;
         fields.forEach(function(field) {
             path.map(function(childPath) {
-                var i = childPath.getName();
                 var lines = print(childPath);
 
                 if (!oneLine) {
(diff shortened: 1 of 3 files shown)


lib/printer.js
```diff
@@ -579,9 +583,9 @@ function genericPrintNoParens(path, options, print) {
         var oneLine = (isTypeAnnotation && len === 1) || len === 0;
         var parts = [oneLine ? "{" : "{\n"];
 
+        var i = 0;
         fields.forEach(function(field) {
             path.map(function(childPath) {
-                var i = childPath.getName();
```
cpojer: because we are iterating over three different fields for `ObjectTypeAnnotation` this counter was wrong and reset for every individual field (`ObjectTypeAnnotation` is a really nasty type). I made it so that it counts the fields in the order they are printed. This fixes additional newlines that were accidentally printed in the object.

lib/printer.js
```diff
@@ -1301,7 +1308,7 @@ function genericPrintNoParens(path, options, print) {
             );
         }
 
-        parts.push(path.call(print, "body"));
+        parts.push(" ", path.call(print, "body"));
```
cpojer: now prints `interface C {}` instead of `interface C{}`

benjamn:
Looks good, thanks!

benjamn:
https://github.com/benjamn/recast/releases/tag/v0.10.20

PR #200, opened 2015-07-28
Add support for printing template expressions.

cpojer:
This adds support for printing template literals.

I did this pretty fast so let me know if I missed anything obvious. 

(Also sneaked in a small code update for printing ObjectExpressions and removed `should` from some tests).

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -585,7 +585,7 @@ function genericPrintNoParens(path, options, print) {
 
         var i = 0;
         fields.forEach(function(field) {
-            path.map(function(childPath) {
+            path.each(function(childPath) {
                 var lines = print(childPath);
 
                 if (!oneLine) {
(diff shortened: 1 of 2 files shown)


lib/printer.js
```diff
@@ -1155,6 +1155,36 @@ function genericPrintNoParens(path, options, print) {
...
+        return n.value.raw;
+
+    case "TemplateLiteral":
+        var expressions = [];
+        var parts = [];
+        var i = 0;
+
+        path.each(function(childPath) {
+            expressions.push(print(childPath));
+        }, "expressions")
```
benjamn: This can just be
```js
var expressions = path.map(print, "expressions");
```

lib/printer.js
```diff
@@ -1155,6 +1155,36 @@ function genericPrintNoParens(path, options, print) {
...
+
+        path.each(function(childPath) {
```
fkling: It seems the index is passed to the `each` callback as second argument, so there is no need for your own loop variable.

benjamn: If that doesn't work, `childPath.getName()` should also be the index of the child in the `.expressions` list.

cpojer:
Simplified the solution. Had some trouble with path.map before which is why I did the manual iteration.

lib/printer.js
```diff
@@ -1155,6 +1155,29 @@ function genericPrintNoParens(path, options, print) {
 
         return concat(parts);
 
+    case "TemplateElement":
+        return n.value.raw;
+
+    case "TemplateLiteral":
+        var expressions = path.map(print, "expressions");
+        var parts = [];
+        path.map(function(childPath) {
```
benjamn: I think it's appropriate to use `path.each` here, since you're not using the resulting array.

lib/printer.js
```diff
@@ -1155,6 +1155,29 @@ function genericPrintNoParens(path, options, print) {
 
         return concat(parts);
 
+    case "TemplateElement":
+        return n.value.raw;
```
benjamn: This technically needs to be `fromString(n.value.raw, options)`, since the rest of the return statements in this function return `Lines` objects.

cpojer:
Updated again! :)

PR #202, opened 2015-07-30
Fix printing class methods

cpojer:
The kind of MethodDefinition changed from init to method in ast-types. This fixes the printer.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1619,7 +1619,7 @@ function printMethod(path, options, print) {
         parts.push("async ");
     }
 
-    if (!kind || kind === "init") {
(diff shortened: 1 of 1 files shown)


benjamn:
Maybe let's preserve backwards compatibility with `kind:"init"`?

cpojer:
I don't think that's necessary – unless you are using an old version of ast-types with a new version of recsat (assuming no one creates nodes manually)?

benjamn:
Right, but `recast.print` is supposed to work with any reasonable AST, not just ones generated by `ast-types`.

cpojer:
Alright, made it backwards compatible :)

PR #207, opened 2015-08-06
[rfc] Respect all line terminators when splitting lines

fkling:
I'm not sure if this is the right approach or if this is needed at all, but my investigation lead me to believe that there is an issue with the `fromString` implementation.

Context: There are problems with using react-docgen with files that use Windows line terminators (see reactjs/react-docgen/issues/11), specifically reading docblock comments.
I added a test which throws an error inside recast: https://gist.github.com/fkling/d31c7729c3c0f521ed39.

It seems that recast doesn't split on all of the line terminators that are defined in the [spec][]. So modified `fromString` and added corresponding tests. Unfortunately this makes it fail for two other tests: https://gist.github.com/fkling/cc043e2afe503d8cb5ba

That lead me to look at the `ExoticWhitespace` test which *looks* like it should actually cover my use case.

Any idea what's going on?

[spec]: http://www.ecma-international.org/ecma-262/5.1/#sec-7.3

### DIFF
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -118,6 +118,9 @@ function countSpaces(spaces, tabWidth) {
 exports.countSpaces = countSpaces;
 
 var leadingSpaceExp = /^\s*/;
+var lineTerminatorSeqExp =
+    /\u000D\u000A|\u000D(?!\u000A)|\u000A|\u2028|\u2029/;
+
 
 /**
  * @param {Object} options - Options object that configures printing.
@@ -137,7 +140,7 @@ function fromString(string, options) {
     if (cacheable && hasOwn.call(fromStringCache, string))
         return fromStringCache[string];
 
-    var lines = new Lines(string.split("\n").map(function(line) {
+    var lines = new Lines(string.split(lineTerminatorSeqExp).map(function(line) {
         var spaces = leadingSpaceExp.exec(line)[0];
         return {
             line: line,
--- a/test/lines.js
+++ b/test/lines.js
@@ -10,6 +10,30 @@ function check(a, b) {
 }
 
 describe("lines", function() {
+
+    describe('line terminators', function() {
+        var source = [
+            'foo;',
+            'bar;',
+        ];
+
+        var terminators = [
+            '\u000A',
+            '\u000D',
+            '\u2028',
+            '\u2029',
+            '\u000D\u000A',
+        ];
+        terminators.forEach(function(t) {
+            it('can handle ' + escape(t) + ' as line terminator', function() {
+                var lines = fromString(source.join(t));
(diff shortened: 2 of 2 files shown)


gonzochic:
Any updates regarding this issue?

benjamn:
Looking into those test failures now.

benjamn:
What about printing? Does Windows need lines to be printed with `\r\n` separators?

benjamn:
Now supporting `options.lineTerminator` to allow configuration when printing: https://github.com/benjamn/recast/commit/371be0a8af6720a78e969cbb21b22b3fe60b3fa5

benjamn:
One more question: is this a breaking change (v0.11-worthy)?

fkling:
> One more question: is this a breaking change (v0.11-worthy)?

If it is really the case that the function doesn't work on all allowed line terminators, then I think it's more of a bug fix imo.

But maybe I'm also totally wrong about this... I don't understand why it breaks the existing tests. Curious to learn what you find out!

benjamn:
Merged with test fixes as part of [0.10.31](https://github.com/benjamn/recast/releases/tag/v0.10.31).

fkling:
Yeah, I missed all the fun part...  :D

fkling:
Oh, I totally misunderstood that exotic white space test. You are adding *all* white space characters. So given that there are 4 single character line terminators, we get 5 lines... makes sense now.

benjamn:
I see from your recent comment that you understand what I'm about to say, but I'm gonna post it anyway for posterity (and sunk-cost nostalgia).

In that `ExoticWhitespace` test, the leading space is a concatenation of literally all space characters:
```js
var source = "";
var spacePattern = /^\s+$/;

for (var i = 0; i < 0xffff; ++i) {
    var ch = String.fromCharCode(i);
    if (spacePattern.test(ch)) {
        source += ch;
    }
}

source += "x";
```
So the `source` ends up looking like
```js
\u0009\u000a\u000b\u000c\u000d\u0020\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u2028\u2029\u202f\u205f\u3000\ufeff\u0078
```
which indeed contains four line terminator sequences, and
```js
"\u0009\u000a\u000b\u000c\u000d\u0020\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u2028\u2029\u202f\u205f\u3000\ufeff\u0078"
  .split(/\u000D\u000A|\u000D(?!\u000A)|\u000A|\u2028|\u2029/).length === 5
```

PR #208, opened 2015-08-07
Add option to print object patterns on one line

kyldvs:
This allows you to print object patterns on one line. The option is put on each node because you generally will not want to configure it globally for all object patterns.

This is important for us when using requires, right now

```
var apple = require('apple');
var {foo, bar} = require('foobar');
var banana = require('banana');
```

Always gets reprinted as

```
var apple = require('apple');
var {
  foo,
  bar
} = require('foobar');
var banana = require('banana');
```

Which isn't really acceptable when you have a long list of requires.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -581,7 +581,7 @@ function genericPrintNoParens(path, options, print) {
             len += n[field].length;
         });
 
-        var oneLine = (isTypeAnnotation && len === 1) || len === 0;
+        var oneLine = n.oneLine || (isTypeAnnotation && len === 1) || len === 0;
         var parts = [oneLine ? "{" : "{\n"];
 
         var i = 0;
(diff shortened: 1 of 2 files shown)


benjamn:
Hmm, I wonder if the default for `ObjectPattern` should be to print everything on one line?

benjamn:
In fact, the default could be one line, and then you could fall back to multi-line printing if the line was too long, something like what I currently do in [`printFunctionParams`](https://github.com/benjamn/recast/blob/cb99299cc574b82a594df73c459e733fbd083ade/lib/printer.js#L1690-L1698).

kyldvs:
I would be a bit concerned about making it length based because even if a require exceeds a few hundred columns I would want to be able to force it onto one line. But in regular code it should not be allowed to get that long. In our codebase we disable line length lints in our imports/requires.

for now I've managed to hack around it though so it's a bit less pressing, http://felix-kling.de/esprima_ast_explorer/#/jndbgQHCE2/16 :)

pieterv:
@benjamn yeah that sounds like the perfect default, that is the primary convention I have observed.

timonag:
Can we have this fixed and merged?

PR #212, opened 2015-08-26
Added parseOptions key, for arbitrary additional options.

benmosher:
Found myself wanting to use `acorn-jsx` as the underlying parser for `recast`, but to activate the JSX plugin, one must call

``` js
acorn.parse(code, { plugins: { jsx: true } })
```

This PR adds a `parseOptions` key which is merged into the normal Recast parse options.

Note: there is a potentially viable alternative in exporting the tab-stripping and tree `Program` modification separately from the `parse` function and allowing callers to call their own parser module directly, and pass the resulting AST into some `recast.format(ast)` function.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -4,6 +4,9 @@ var defaults = {
     // recast.parse as options.esprima.
     esprima: require("esprima-fb"),
 
+    // Additional options passed to the parse module (i.e. Acorn plugins)
+    parseOptions: null,
+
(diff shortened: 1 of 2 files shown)


demoneaux:
I think this PR is not required as you can just pass a custom parser into recast via [`options.parser`](https://github.com/benjamn/recast/blob/9b32ded58db4538b644d59a156a0a95959b147a1/lib/options.js#L2-L5).

PR #218, opened 2015-10-20
Always wrap arrow functions in parenthesis if the only param isn’t an identifier.

### LINKED ISSUE #217: Missing parenthesis on async arrow function where the first argument is an array destructuring
Input:

```javascript
([a, b]) => {}
```

codemod (just puts `async = true`)

```javascript
module.exports = function(file, api) {
  const j = api.jscodeshift;
  const {expression, statement, statements} = j.template;

  return j(file.source)
    .find(j.ArrowFunctionExpression)
  	.forEach(p => p.value.async = true)
    .toSource();
};
```

result:

```javascript
async [a, b] => {}
```

but this doesn't parse. Should be

```javascript
async ([a, b]) => {}
```

screen shot 2015-10-18 at 8 17 57 pm

cpojer:
Fixes #217

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -282,8 +282,7 @@ function genericPrintNoParens(path, options, print) {
         if (
             n.params.length === 1 &&
             !n.rest &&
-            n.params[0].type !== 'SpreadElementPattern' &&
-            n.params[0].type !== 'RestElement'
+            n.params[0].type === 'Identifier'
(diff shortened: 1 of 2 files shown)


benjamn:
Published [v0.10.34](https://github.com/benjamn/recast/releases/tag/v0.10.34) to NPM.

PR #227, opened 2015-11-04
Print class decorators (not only class methods decorators)

shumov:
Now it can print by ast
```javascript
@decorator
class Some {
  method() {
  }
}
```

not only
```javascript
class Some {
  @decorator
  method() {
  }
}
```
as yesterday


benjamn:
Exactly as I would have done it. Thanks!

shumov:
Awesome! )

Issue #228, opened 2015-11-04
Add option to control whether multiline properties in Objects are pretty printed with blank lines


asselin:
Currently, multiline properties in Objects are pretty printed with blank lines before/after them.  For example:
````
        $stateProvider
            .state('map', {
                parent: 'root',
                url: '/map',

                views: {
                    'content@app': {
                        templateUrl: 'app/map/map.html',
                        controller: 'MapController',
                        controllerAs: 'vm'
                    }
                }
            });
````

Add an option to control whether the extra blank line is inserted.  For example, setting to false would produce:
````
        $stateProvider
            .state('map', {
                parent: 'root',
                url: '/map',
                views: {
                    'content@app': {
                        templateUrl: 'app/map/map.html',
                        controller: 'MapController',
                        controllerAs: 'vm'
                    }
                }
            });
````

benjamn:
There are a couple of distinct considerations here:
1. how the pretty-printer should print entirely new code (related: #226)
2. how object properties should be "reprinted" when only some of the properties have changed

Re: 2., I think there's an opportunity here to do what [`printStatementSequence`](https://github.com/benjamn/recast/blob/060b6fc9df7956a6f39df4444a04740c8e57621d/lib/printer.js#L1554-L1602) does for statements, so whitespace between pre-existing object properties is preserved, rather than blown away. Then you wouldn't need to rely on a dedicated printer option to achieve the goal of preserving existing whitespace.

Re: 1., I like to avoid accumulating special-purpose options if possible, so I think it's worth asking if we should just change the printer so that it always puts newlines around multi-line properties (when printing entirely new code). That's [how it works for statements](https://github.com/benjamn/recast/blob/060b6fc9df7956a6f39df4444a04740c8e57621d/lib/printer.js#L1573) currently. Unless you can think of situations where you would really want there to be no newlines around properties (throughout the entire file), then let's just change the printer behavior to prefer spaces and not introduce a new option?

asselin:
@benjamn I agree on your 2nd point.  For the first point, this is a matter of preference and possibly coding standards, so I firmly believe it needs to be an option.  However, I do agree with your point about not making a ton of special purpose options.

I looked at .eslintrc, and didn't see a rule related to this-- if there was, I'd say this could be the first option implemented in the formatter.  I suppose it could still be part of that infrastructure, and it could be a recast extension.  Any suggestions on how you'd like to see options like this implemented?

qimingweng:
[...]

PR #229, opened 2015-11-04
Add lineBreaksInObjects option to control whether pretty-printed Objects have line breaks between multiline properties

asselin:
This is for issue #228

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -64,6 +64,10 @@ var defaults = {
     // array expressions, functions calls and function definitions pass true
     // for this option.
     trailingComma: false,
+
+    // If you want line breaks before and after multiline properties in objects,
+    // pass true for this option.
+    lineBreaksInObjects: true,
 }, hasOwn = defaults.hasOwnProperty;
 
 // Copy options and fill in default values.
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -612,8 +612,8 @@ function genericPrintNoParens(path, options, print) {
                 if (i < len - 1) {
                     // Add an extra line break if the previous object property
                     // had a multi-line value.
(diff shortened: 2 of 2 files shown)


jcuffe:
@benjamn Just found this excellent package, but was confused with the default behavior made configurable here.

I'm using this library to add nested properties to an object in one of our central modules, and the introduced newlines result in a large diff. I'm using `patch-package` to get around this locally, but would love to see this PR merged 🙏 

As an aside, I'm curious to know if I'm using the library incorrectly since all properties of the object are being reprinted with the generic printer, while formatting in sibling keys is preserved as expected. Here's the gist of it:

```
// Load and parse the file
// ...

// Find the job handlers object inside module exports
const exportStatement = _.find(ast.program.body, ['expression.left.object.name', 'module']);
const exportProps = _.get(exportStatement, 'expression.right.properties');
const jobHandlersObj = _.find(exportProps, ['key.name', 'jobHandlers']);
const jobHandlerProps = _.get(jobHandlersObj, 'value.properties');

// Check for an existing require to prevent duplication
// ...

// Add a getter to require the new job module
jobHandlerProps.push(
  b.property(
    'get',
    b.identifier(moduleName),
    b.functionExpression(
      null,
      [],
      b.blockStatement([
        b.returnStatement(
          b.callExpression(b.identifier('require'), [b.literal(PATHS.loaderRequire(moduleName))]),
        ),
      ]),
    ),
  ),
);

// Write the modified source to the loader file
await fs.writeFile(PATHS.LOADER, recast.print(ast).
```

Thanks for your work maintaining this! I was pleasantly surprised to find this package already in my dependency tree, powering other codemod tools we use in our project.

Issue #230, opened 2015-11-04
Indentation level is wrong for nodes that cross multiple lines, where the lines are at different indent levels


asselin:
When there is a node who's original source code crosses multiple lines, and underneath it, something needs to be pretty printed, the indentation level should match the last line of the original source code.  For example, given:

````
        $stateProvider
            .state('map', {
                foo: 'bar'
            });
````
Currently, if that object has to be pretty printed (such as when adding a property to it), the result will be:

````
        $stateProvider
            .state('map', {
            foo: 'bar'
        });
````

The reason is that the indentation level being used is for the the first line ($stateprovider) instead of the 2nd line (.state).

benjamn:
I think you're right, and this would be a worthwhile improvement.

Without thinking about it in more depth, I'm not sure what the new logic should look like, but I'm pretty sure it will involve some changes [here](https://github.com/benjamn/recast/blob/060b6fc9df7956a6f39df4444a04740c8e57621d/lib/parser.js#L119-L122).

asselin:
@benjamn Well, I solved it a different way... take a look at my PR and let me know what you think.

PR #231, opened 2015-11-04
Fix indentation to handle nodes that cross multiple lines, and the lines are at different indent levels.

asselin:
This is for issue #230

### DIFF
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -195,7 +195,7 @@ exports.getReprinter = function(path) {
             var newLines = print(
                 reprint.newPath,
                 needToPrintNewPathWithComments
-            ).indentTail(oldNode.loc.indent);
+            ).indentTail(oldNode.loc.lines.getIndentAt(oldNode.loc.start.line));
 
             var needsTrailingSpace =
                 riskyAdjoiningCharExp.test(lines.charAt(oldNode.loc.end));


benjamn:
Want to add some tests?

asselin:
The tests all fail on Windows (I suspect probably a line ending issue).  I'll take a look at that first, and then at tests.  Any suggestions on what you'd like to see in the way of tests?

asselin:
So yes, \n vs \r\n is what's causing the tests to fail on Windows.  What's your preference on how this should be fixed?  A couple of options are:

1. Add the lineTerminator option to every print call.  e.g.

````
var printOptions = {
    lineTerminator: "\n"
};
recast.print(dup.body, printOptions).code
````

2. Change every place that concatenates lines together to create "expected" to use the OS line terminator.  e.g.

````
var eol = require("os").EOL;
var code = trailing.join(eol);
````

Personally, I think option 2 is a better choice.  It's cleaner in the code, and to me it more closely tests what the library is supposed to be doing.

@benjamn What's your opinion?

benjamn:
I agree that option 2 is better, since it exercises more of the parsing/printing logic. Can Travis run tests on Windows automatically?

asselin:
Looks like Travis supports only Mac and Linux.  https://github.com/travis-ci/travis-ci/issues/216

I'll submit a PR for option 2 later today

benjamn:
That would be great, thanks!

eventualbuddha:
Closing since this never got tests added.

PR #232, opened 2015-11-06
Fix the unit tests so that they don't assume the OS they're being run on uses \n as the line terminator

asselin:


benjamn:
Thanks so much for taking care of this!

asselin:
No problem.  I'll work on the tests for the other PR as soon as I get some time.

PR #233, opened 2015-11-11
Enable ES2015+ syntax in Recast implementation via Babel 6

benjamn:
It's pretty ironic that Recast has supported ES2016+ syntax in transformed code for more than two years, but its own implementation is still plain ES5! I think it's time Recast caught up with itself. Selfishly, I also know the new syntax is going to make maintaining this project much more enjoyable for me.


jamestalmage:
@benjamn 
My only thought would be to move the old contents of `index.js` to `main.js`, and then have this in index.js:

```js
require('babel-core/register');
require('./main');
```

Then people can set the require hook the way they want it for their own lib and just do:

```js
require('recast/main');
```

To bypass your babel hook if they know they have already loaded up Babel or are running on some future Node version that has all the ES6 features natively.

benjamn:
@jamestalmage that's a good idea

benjamn:
I'm only doing `require("babel-core/register")` in the tests, though… did you mean `require("babel-polyfill")`?

jamestalmage:
oops, saw `index.js` and skipped right over which folder it was in.

I'm assuming you are going to publish transpiled results then?

benjamn:
Yep, I'll publish the `lib/` directory in addition to the `src/` directory (trying to follow [my own advice](http://benjamn.github.io/empirenode-2015/#/73)).

benjamn:
Dropping support for Node versions earlier than 4 turned out to be more effective on this front.

PR #236, opened 2015-11-25
Fix printing implements keyword

juliankrispel:
Only print implements keyword if implements property has more than 0 items

Basically printing a `ClassExpression` adds an implements keyword. I'm guessing that this check here should really also check if the implements array is empty and that's why I get that result, because the default for `ClassExpression.implements` is `[]`.

Here's how this looks like:

http://astexplorer.net/#/mNvVofHYVs/1

cc @benjamn

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1172,7 +1172,7 @@ function genericPrintNoParens(path, options, print) {
             );
         }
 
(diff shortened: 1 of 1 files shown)


juliankrispel:
thanks @benjamn that was quick

PR #244, opened 2016-01-06
Change build badge to svg.

knpwrs:
Makes a difference on retina displays. The gitter badge is already svg anyway so I assume you're cool with using them.

image


benjamn:
Nice catch!

Issue #178, opened 2015-04-16
esprima-fb → esprima?


mathiasbynens:
https://github.com/benjamn/recast/blob/24ebc0e110aed78326fc1278d888912c4a53cb7b/package.json#L28

Now that esprima-fb is being deprecated (https://github.com/facebook/esprima/issues/111), will recast switch to regular Esprima instead?

benjamn:
Yes, I think I would like to run the test suite using both https://github.com/jquery/esprima and https://github.com/marijnh/acorn. Any other suggestions?

mathiasbynens:
There’s also https://github.com/eslint/espree.

danieljuhl:
babel-core?

cpojer:
I think https://github.com/babel/acorn-babel would be a good contender, despite the README saying it is not supported outside of babel. Babel is quickly becoming a transpiler standard.

fkling:
@cpojer: I don't think that project is kept up-to-date anymore. The acorn fork/extension seems to be kept inside babel-code.

fkling:
Is there any progress on supporting different parsers? If you are busy, maybe you could outline what needs to be done so that's easier for others to help?

benjamn:
In principle, `recast.parse(source, { esprima: require("any/parser/you/like") })` should work already for any sufficiently Esprima-like parser module (including Acorn AFAIK), so it should be just a matter of actually running tests using different parsers to provide more confidence there?

benjamn:
Specifically, I would add other parsers to devDependencies in package.json, then allow the test suite to `require` a different parser based on an environment variable like `process.env.PARSER_MODULE_ID`, then add a `npm` script called `test-all-parsers` that invokes (at least the parser-related) tests for every parser. That comprehensive test should be what runs in Travis CI, though it's fine if the ordinary `npm test` command just uses the default parser.

This testing might reveal that more pretty-printer cases need to be added to `lib/printer.js`, but that work should be minimal.

nene:
It's been a while and esprima-fb is still deprecated. Should Recast also be considered as deprecated?

Like, I'm trying to parse this code:

```js
function fn({foo={}} = {}) { }
```

...and Recast simply blows up because esprima-fb is unable to handle it. Official Esprima parses it just fine though.

fkling:
@nene: you can always pass another parser to recast via the `esprima` option. But I agree that recast should upgrade to another default parser or not provide a default at all (which would be a breaking change though).

[...]

PR #247, opened 2016-01-14
Attempt to use babylon as the default parser.

benjamn:
This pull request currently has a lot of test failures, but perhaps it's not as far from working as it seems.

Previous discussion of the default parser choice: #178

To be clear: `recast` works perfectly well with [`babylon` as a custom parser](https://github.com/benjamn/recast/blob/892a12e2b3ef5313ab6af135c309c2fad59b0ef7/test/babylon.js#L29-L35), there are just a lot of existing tests that implicitly depend on `esprima(-fb)` parsing quirks.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -1,8 +1,35 @@
+var babylon = require("babylon");
+var babylonOptions = {
+    sourceType: 'module',
+    allowImportExportEverywhere: false,
+    allowReturnOutsideFunction: false,
+    plugins: [
+        'asyncFunctions',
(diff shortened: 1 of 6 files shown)


benjamn:
This is easier now that
```js
const ast = recast.parse(code, {
  parser: require("recast/parsers/babylon")
});
```
works (https://github.com/benjamn/recast/pull/472). Given that, I think we'll keep Esprima as the default for now, if only for backwards compatibility.

PR #249, opened 2016-01-15
Fixes #248

### LINKED ISSUE #248: Invalid pretty-print for arrow function with single argument and a flow type definition
I ran the [bind to arrow function codemod](https://github.com/jhgg/js-transforms/blob/master/call-expression-bind-this-to-arrow-function-expression.js) on react-native and it transformed:

```diff
- onResponderTerminate: function(e: SyntheticEvent) {
+ onResponderTerminate: e: SyntheticEvent => {
```

As you can see, it's missing a parenthesis.

vjeux:
Tested on the codemod and it works :)

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -339,7 +339,8 @@ function genericPrintNoParens(path, options, print) {
         if (
             n.params.length === 1 &&
             !n.rest &&
-            n.params[0].type === 'Identifier'
+            n.params[0].type === 'Identifier' &&
+            !n.params[0].typeAnnotation
         ) {
             parts.push(path.call(print, "params", 0));
         } else {


benjamn:
Makes sense to me!

Just so you know, there was a minor version bump recently (to 0.11), which changed the default parser from `esprima-fb` (which supports type annotations) to plain `esprima` (which only supports ES2015 stuff), so in order to use this change you'll probably need to upgrade and then also provide your own parser via `recast.parse(source, { parser: ... })`.

vjeux:
Cool thanks! I'm using jscodeshift to test this which probably sets the proper parser :)

@felixkling also suggested that I write a test, which I will do

PR #250, opened 2016-01-15
Add test for arrow function parenthesis with one typed argument

vjeux:


vjeux:
See #249 for the actual code :)

PR #252, opened 2016-02-11
Print rest params in FunctionTypeAnnotations

garetht:
Although the `FunctionTypeAnnotation` specifies a [`rest` field](https://github.com/benjamn/ast-types/blob/master/def/fb-harmony.js#L162), the printer does not account for the existence of this field and does not correctly print it. This pull request adds the ability to print the `rest` field as well as adds a brief test.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1300,9 +1300,14 @@ function genericPrintNoParens(path, options, print) {
             parts.push(": ");
         }
 
+        var params = path.map(print, "params");
+        if (n.rest) {
+            params.push(concat(["...", path.call(print, "rest")]));
+        }
+
         parts.push(
             "(",
-            fromString(", ").join(path.map(print, "params")),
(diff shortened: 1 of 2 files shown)


lib/printer.js
```diff
@@ -1300,9 +1300,14 @@ function genericPrintNoParens(path, options, print) {
             parts.push(": ");
         }
 
+        var params = path.map(print, "params");
+        if (n.rest) {
+            params.push(concat(["...", path.call(print, "rest")]));
```
benjamn: The `concat([...])` is not necessary here; `params.push("...", path.call(print, "rest"))` should work.

vicapow:
bump

eventualbuddha:
This appears to have been fixed by https://github.com/benjamn/recast/commit/7e7dbc9766f71050985ed8d4bb9449c8c390741f.

Issue #191, opened 2015-07-02
Trailing line comments are being turned into leading comments...


zertosh:
but only line comments, not block comments.

```js
var recast = require('recast');

var ast = recast.parse([
  'module.exports = {};',
  '// trailing line comment',
  '/**',
  ' * trailing block comment',
  ' */'
].join('\n'));

recast.visit(ast, {
  visitNode: function(path) {
    this.traverse(path);
    path.value.original = null;
  }
});

console.log(recast.print(ast).code);
```

Actual:
```
// trailing line comment
module.exports = {};
/**
 * trailing block comment
 */;
```

Expected:
```
module.exports = {};
// trailing line comment
/**
 * trailing block comment
 */;
```

benjamn:
I believe this behavior is a consequence of [an imperfect decision](https://github.com/benjamn/recast/blob/c240839a2c729443875a8e67eebf12ad8319814d/lib/comments.js#L340-L341) that I made a while back to reclassify trailing line comments as leading comments when pretty printing, since printing trailing comments (especially end-of-line ones) is tricky.

<em>Please keep in mind that this bug only manifests itself when `recast.print` falls back to pretty printing the node that the comments are attached to. Otherwise, `recast.print` does a pretty good job of preserving tricky formatting. For example, if you change only the text of the comments, they will be reprinted in place, without triggering the pretty printer. Setting `node.original = null` is a reliable way to force the pretty printer to kick in, so I understand why you used it for this reproduction. I just hope the scope of the problem is clear: this bug reflects a limitation of the pretty printer, and `recast.print` avoids the pretty printer as much as possible.</em>

Let me explain what's tricky, and you can tell me if you see a less-bad solution?

Imagine you have a multi-line function call with comments for the arguments:
```js
addNumbers(
  a, // the first number
  b, // the second number
  c // the third number
);
```

All of these are trailing comments, and [the comment attachment code](https://github.com/benjamn/recast/blob/c240839a2c729443875a8e67eebf12ad8319814d/lib/comments.js#L120) gets that right.

[...]

PR #259, opened 2016-02-19
Print trailing comments on statements as trailing

dgreensp:
Doing so doesn't break any existing tests.

@benjamn: This fixes a test case I have in my downstream project, and the provided test case.  If it doesn't have any unforeseen consequences, it may be a fix for #191 for most purposes.

### DIFF
--- a/lib/comments.js
+++ b/lib/comments.js
@@ -337,7 +337,8 @@ exports.printComments = function(path, print) {
         var leading = types.getFieldValue(comment, "leading");
         var trailing = types.getFieldValue(comment, "trailing");
 
-        if (leading || (trailing && !(comment.type === "Block" ||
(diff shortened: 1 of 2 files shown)


benjamn:
I think this is probably more helpful than harmful, so I'm happy to merge it without completely understanding the consequences. In my experience the only way to approach comment handling improvements is write a bunch of tests inspired by the output of real codemods. Speculating about the consequences without concrete examples just doesn't work. If this change causes problems, then we'll write some more tests and figure out a way to add a little bit more nuance to the rules.

If it turns out this needs more nuance, we could check that the node is not only a `Statement` but a direct child of a `BlockStatement` or `Program`—in other words, it has to be part of a list of `Statements`s, and not the `.consequent` of an `IfStatement` or something like that.

PR #263, opened 2016-03-02
Preserve leading and trailing whitespace around string children of JSXElements

### LINKED ISSUE #246: Leading/trailing whitespace in JSXElements are stripped by design
Ref b50620d

The recast printer deliberately removes leading/trailing whitespace from the children of JSXElements, which breaks my JSX code.

`<strong> hi</strong>`

becomes:

`<strong>hi</strong>`

What is the reasoning behind remove the trailing/leading whitespace?

Possibly related to #213, even though that only mentions it going wrong for `prettyPrint`, and this is something that should always go wrong.

ahutchings:
I'm not sure what the intention was behind the whitespace-stripping logic I removed. If there is a specific case that it was meant to handle, let me know and I can update this pull request with a test for it.

Fixes #246.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1023,11 +1023,7 @@ function genericPrintNoParens(path, options, print) {
 
                 if (namedTypes.Literal.check(child) &&
                     typeof child.value === "string") {
-                    if (/\S/.test(child.value)) {
-                        return child.value.replace(/^\s+|\s+$/g, "");
-                    } else if (/\n/.test(child.value)) {
-                        return "\n";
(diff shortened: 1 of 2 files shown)


benjamn:
If I recall correctly, that stripping code made JSX conform a bit better to coding conventions in Facebook's internal codebase, but I can see how it might mess up legitimate code that follows different conventions. Better to avoid special cases. Thanks for the PR!

ahutchings:
Thanks for merging! Would you be willing to cut a release with this change?

PR #284, opened 2016-06-09
Fix #283 - avoid unnecessary parens.

### LINKED ISSUE #283: Extra parens around ReturnStatement with SequenceExpression.
Given this test case:

```js
#!/usr/bin/env node

var recast = require("recast");

var ast = recast.parse(
`function foo() {
  return 1, 2;
}
`);

recast.visit(ast);

process.stdout.write(recast.print(ast).code);
```

I would expect the output to be:

```js
function foo() {
  return 1, 2;
}
```

However, the actual output includes extra parentheses around the SequenceExpression:

```js
function foo() {
  return (1, 2);
}
```

mbostock:


benjamn:
Thanks! I'm away from my computer now but I'll publish this later today or tomorrow morning.

PR #285, opened 2016-06-11
Add support for ES7 exponentiation operator

nene:
The change itself is trivial - just adding `**` operator to PRECEDENCE table.

I'm not really sure about the proper way to test it. Added a test to Babylon test suite, which covers the problem that I was running into. Though it might seem more logical to include it to the general Parens test suite, but Esprima doesn't yet support this operator yet.

For reference: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -374,7 +374,7 @@ var PRECEDENCE = {};
  ["<", ">", "<=", ">=", "in", "instanceof"],
  [">>", "<<", ">>>"],
  ["+", "-"],
- ["*", "/", "%"]
+ ["*", "/", "%", "**"]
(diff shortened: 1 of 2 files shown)


nene:
BTW. Recast contains `.editorconfig` which defines `indent_size = 4` but some files in the repository use 2 spaces for indentation. So my editor tried to force me into using 4 spaces while editing the file with 2-space indentation. Fiddled for a while with my editor configuration... thinking there must be something wrong with it that it wants to indent by 4 spaces in a file the obviously has 2-space indentation.

Like really... this is just silly.

benjamn:
Thanks!

PR #288, opened 2016-06-13
Add singleArgumentArrowParenthesis option

GoodForOneFare:
For users with eslint configured with `arrow-parens: always`.


benjamn:
I like this idea, but can we just call it `options.arrowParensAlways`?

GoodForOneFare:
@benjamn Good call - done!

GoodForOneFare:
@benjamn anything else I can do to push this PR along?

GoodForOneFare:
Thank you! 😄

PR #291, opened 2016-06-23
Add `flowObjectCommas` option

keyanzhang:
Currently there are 2 supported syntaxes (`,` and `;`) in Flow Object Types. The use of commas is in line with the more popular style and matches how objects are defined in JS, making it a bit more natural to write. Recently we added a lint rule that enforces the use of commas, so it would great to have this option in recast as well. Thanks! 😃 

cc @cpojer

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -68,6 +68,11 @@ var defaults = {
     // If you want parenthesis to wrap single-argument arrow function parameter
     // lists, pass true for this option.
     arrowParensAlways: false,
+
+    // There are 2 supported syntaxes (`,` and `;`) in Flow Object Types;
+    // The use of commas is in line with the more popular style and matches
+    // how objects are defined in JS, making it a bit more natural to write.
+    flowUsesCommas: true,
 }, hasOwn = defaults.hasOwnProperty;
 
 // Copy options and fill in default values.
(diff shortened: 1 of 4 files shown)


keyanzhang:
Also, seems like it currently prints single line indexers with an extra comma (`{[key: string]: number,}`) in the end. It's probably related to the way we handle trailingCommas [here](https://github.com/benjamn/recast/blob/master/lib/printer.js#L608). I'll push a fix when https://github.com/benjamn/ast-types/pull/156 gets reviewed/merged -- thanks!

lib/options.js
```diff
@@ -64,6 +64,11 @@ var defaults = {
     // array expressions, functions calls and function definitions pass true
     // for this option.
     trailingComma: false,
+
+    // There are 2 supported syntaxes (`,` and `;`) in Flow Object Types;
+    // The use of commas is in line with the more popular style and matches
+    // how objects are defined in JS, making it a bit more natural to write.
+    flowUsesCommas: true,
```
keyanzhang: Oh actually, would it be better to change the default to `false` to avoid having a breaking change? @benjamn

benjamn: Let's stick with comma-by-default.

benjamn:
Can we call the option `flowObjectCommas` instead of `flowUsesCommas`?

keyanzhang:
Just renamed. Thanks!

benjamn:
Rebase?

keyanzhang:
@benjamn sorry didn't notice the merge conflict :)

PR #293, opened 2016-06-28
Print parens for nullable union/intersection types

keyanzhang:
Add parens for types like `?(string | number)`. Thanks!

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -192,8 +192,8 @@ FPp.needsParens = function(assumeExpressionContext) {
         return false;
     }
 
-    // Only expressions need parentheses.
-    if (!n.Expression.check(node)) {
+    // Only statements don't need parentheses.
+    if (n.Statement.check(node)) {
         return false;
     }
 
@@ -288,6 +288,10 @@ FPp.needsParens = function(assumeExpressionContext) {
         default:
             return false;
         }
+
+    case "IntersectionTypeAnnotation":
+    case "UnionTypeAnnotation":
+        return parent.type === "NullableTypeAnnotation";
 
     case "Literal":
         return parent.type === "MemberExpression"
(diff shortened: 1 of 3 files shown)


benjamn:
Would it make sense to add this logic to [`FastPath.prototype.needsParens`](https://github.com/benjamn/recast/blob/5c6bf309e7d5bb7618e04b3fd4b07010537b1c0f/lib/fast-path.js#L179)?

keyanzhang:
@benjamn cool, should I just call `maybeAddParens` then?

benjamn:
You shouldn't have to call `maybeAddParens` explicitly, since the `print` function in `genericPrintNoParens` goes through all that logic for you. That's my intention, at least.

If I'm right, you should only have to `return true` from `FastPath#needsParens` when the node is a `UnionTypeAnnotation` or `IntersectionTypeAnnotation` and the parent node is a `NullableTypeAnnotation`.

keyanzhang:
Thanks @benjamn. I added 
```js
    case "IntersectionTypeAnnotation":
    case "UnionTypeAnnotation":
        return parent.type === "NullableTypeAnnotation";
```
as new top-level cases in `#needsParens` but seems these cases are not reached. Am I missing something here?

My new `NullableTypeAnnotation` case in `printer`:
```js
    case "NullableTypeAnnotation":
        return concat([
            "?",
            path.call(print, "typeAnnotation")
        ]);
```

keyanzhang:
I added a diff (https://github.com/benjamn/recast/pull/293/commits/bf190cdb8239e387398520d50ab5e1762bf6198f) to show my failed attempt. Thanks!

benjamn:
[Here](https://github.com/benjamn/recast/blob/1c9725db83e8fe15ca9b2671bdc0a6f52fb74f79/lib/fast-path.js#L195-L198)'s the culprit. Can you change that check to exclude `Statement` nodes rather than allowing only `Expression` nodes?

keyanzhang:
Fixed, thanks!

keyanzhang:
When you have time, could you also bump the `ast-types` dep to `0.8.18` while releasing a new version? Thanks! 😃

PR #299, opened 2016-07-07
Babel 6 changes + flow

hzoo:
Depends on https://github.com/benjamn/ast-types/pull/162 so can change the package.json after it's merged

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -225,9 +225,18 @@ function genericPrintNoParens(path, options, print) {
         return path.call(print, "program");
 
     case "Program":
-        return path.call(function(bodyPath) {
+        // Babel 6
+        if (n.directives) {
+            path.each(function(childPath) {
+                parts.push(print(childPath), ";\n");
+            }, "directives");
+        }
+
+        parts.push(path.call(function(bodyPath) {
             return printStatementSequence(bodyPath, options, print);
-        }, "body");
+        }, "body"));
+
+        return concat(parts);
 
     case "Noop": // Babel extension.
     case "EmptyStatement":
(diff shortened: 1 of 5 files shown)


lib/printer.js
```diff
@@ -225,9 +225,15 @@ function genericPrintNoParens(path, options, print) {
         return path.call(print, "program");
 
     case "Program":
-        return path.call(function(bodyPath) {
+        // if (n.directives) {
```
hzoo: I need help with directives - I think we need something like `printStatementSequence`? Just need to loop over the directives array

lib/printer.js
```diff
@@ -646,6 +656,43 @@ function genericPrintNoParens(path, options, print) {
 
         return concat(parts);
 
+    // fallthrough into ObjectMethod
+    case "ClassMethod": // Babel 6
+        if (n.static) {
+            parts.push("static ");
+        }
+
+        // fallthrough into ObjectMethod
```
hzoo: This is the same as an ObjectMethod except you can have static at the beginning - not sure if we want to do it like this or just move the static check into object method anyway

lib/printer.js
```diff
@@ -698,17 +745,27 @@ function genericPrintNoParens(path, options, print) {
     case "Super":
         return fromString("super");
 
-    case "BooleanLiteral": // Babel 6 Literal split
+    case "NullLiteral": // Babel 6 Literal split
+        return fromString("null");
```
hzoo: Null doesn't have a value since its already null

lib/printer.js
```diff
@@ -698,17 +745,27 @@ function genericPrintNoParens(path, options, print) {
...
+
     case "RegExpLiteral": // Babel 6 Literal split
+        return fromString(n.extra.raw);
```
hzoo: RegExp doesn't have a value eithr

lib/printer.js
```diff
@@ -1674,7 +1731,7 @@ function printArgumentsList(path, options, print) {
 
 function printFunctionParams(path, options, print) {
     var fun = path.getValue();
-    namedTypes.Function.assert(fun);
+    // namedTypes.Function.assert(fun);
```
hzoo: Need to fix this somewhere for ObjectMethod I think

PR #302, opened 2016-07-18
Fix assumption of declaration property on export

BrendanAnnable:
Not all export declaration nodes have a `declaration` property. e.g. `export { MyObject, MyOtherObject}` has no `declaration` but an array of `specifiers`

When printing it will crash with an `TypeError: Cannot read property 'decorators' of null` error.


benjamn:
Thanks @BrendanAnnable!

PR #303, opened 2016-07-18
Fix export trailing semicolon

BrendanAnnable:
When printing exported class or function declarations, recast currently prints an unnecessary trailing semicolon. This PR provides a fix to omit them.

These are the only two scenarios I could think of that would not require a semicolon.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1740,7 +1740,12 @@ function printExportDeclaration(path, options, print) {
 
     var lines = concat(parts);
 
-    if (lastNonSpaceCharacter(lines) !== ";") {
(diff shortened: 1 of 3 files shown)


BrendanAnnable:
Not sure how you'd like to handle tests in this case, I'll leave it up to you 👍

apexearth:
This went into 0.11.11 and has caused functional changes which has negatively affected **mohebifar/lebab**.

benjamn:
@apexearth I'm willing to revert this change, unless the tests are easily fixable on your side.

BrendanAnnable:
Just to wrap up, this has been fixed downstream in lebab now https://github.com/mohebifar/lebab/pull/160

PR #304, opened 2016-07-20
Update to ast-types 0.8.18

fkling:
... to include `TypeParameter` information (https://github.com/benjamn/ast-types/commit/d5ce0400511bdf6b063c13728c4fa09edd9e122d). See also https://github.com/flowtype/flow-codemod/issues/6#issuecomment-233738594 .

This would also require a new (patch) release of recast.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1424,6 +1424,28 @@ function genericPrintNoParens(path, options, print) {
             fromString(", ").join(path.map(print, "params")),
             ">"
         ]);
+    case "TypeParameter":
+        switch (n.variance) {
+            case 'plus':
+                parts.push('+');
+                break;
+            case 'minus':
+                parts.push('-');
+                break;
+            default:
+        }
+
+        parts.push(path.call(print, 'name'));
+
+        if (n.bound) {
+            parts.push(path.call(print, 'bound'));
+        }
+
+        if (n['default']) {
+            parts.push('=', path.call(print, 'default'));
+        }
+
+        return concat(parts);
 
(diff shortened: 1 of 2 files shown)


fkling:
Mmh, I guess the printer needs to be updated? I'll look into it.

fkling:
Updated the printer as well. Tested the print result in the node REPL:

```js
var n = {
    "type": "TypeParameter",
    "name": "X",
    "bound": {
        "type": "TypeAnnotation",
        "typeAnnotation": {
            "type": "GenericTypeAnnotation",
            "id": {
                "type": "Identifier",
                "name": "T",
                "typeAnnotation": null,
                "optional": false
            },
            "typeParameters": null
        }
    },
    "variance": "minus",
    "default": {
        "type": "NumberTypeAnnotation",
    }
};
recast.print(n); // PrintResult { code: '-X: T=number' }
```

cc @jeffmo, @keyanzhang regarding the generated code. I don't know how `NullTypeAnnotation` should be printed (if at all), it doesn't seem to be used by any other nodes.

jeffmo:
`NullTypeAnnotation` is the `null` annotations:

```js
var a: null = null;
```

fkling:
Ah OK. And it's not referenced anywhere else in the definition because it simply extends `Type`. Alright, will update the PR.

fkling:
Ping? Without this, jscodeshift is unusable on files that use the latest flow syntax, @benjamn. 😞

benjamn:
Done! Thanks for the bump. https://github.com/benjamn/recast/releases/tag/v0.11.11

PR #311, opened 2016-08-26
Guess line terminators - fixes #269

### LINKED ISSUE #269: Make `lineTerminator` inferred based on the source (like `tabWidth`)
Hi.

Recast chooses the `lineTerminator` to use based on the default end-of-line marker of the host OS.
See this line:
https://github.com/benjamn/recast/blob/539d044ef2d295878e056f6b5c9648a7c05f9823/lib/options.js#L21

This can cause every single line in the original file appear changed if you're using Windows to run a codemod on a file that was originally written in Mac\Linux, or vice-versa.
(As a Windows user, most of the open source-code I depend on uses `\n` and not `\r\n` because it was not written on Windows, so this is quite a common scenario for me).

Is it possible to make recast infer the original EOL marker based on the original file, like it does for `tabWidth`?

Thank you!

Full disclosure: I didn't run `recast` directly but rather via `jscodeshift`, but I'm assuming this is still the relevant repo to address this issue.

Please LMK if you need any further clarifications from me.
Would be happy to take a swing at submitting a PR with some guidance as to how this could be achieved.

peet:
Uses the most commonly found line ending in the source as the terminators for the whole output. Hopefully it handles splitting and joining bunches of lines properly too.

Note: I have tested this relatively limited, with [mohebifar/lebab](https://github.com/mohebifar/lebab). Could do with being dropped into something more complicated for a more thorough test.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,2 +1,3 @@
 [*.js]
 indent_style = space
+indent_size = 4
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -122,7 +122,7 @@ var leadingSpaceExp = /^\s*/;
 
 // As specified here: http://www.ecma-international.org/ecma-262/6.0/#sec-line-terminators
 var lineTerminatorSeqExp =
-    /\u000D\u000A|\u000D(?!\u000A)|\u000A|\u2028|\u2029/;
+    /\u000D\u000A|\u000D(?!\u000A)|\u000A|\u2028|\u2029/g;
 
 /**
  * @param {Object} options - Options object that configures printing.
@@ -143,6 +143,7 @@ function fromString(string, options) {
     if (cacheable && hasOwn.call(fromStringCache, string))
         return fromStringCache[string];
 
+    var terminators = string.match(lineTerminatorSeqExp) || [];
     var lines = new Lines(string.split(lineTerminatorSeqExp).map(function(line) {
         var spaces = leadingSpaceExp.exec(line)[0];
         return {
@@ -157,6 +158,8 @@ function fromString(string, options) {
 
     if (cacheable)
         fromStringCache[string] = lines;
+
+    getSecret(lines).terminators = terminators;
 
     return lines;
 }
@@ -330,6 +333,8 @@ Lp.stripMargin = function(width, skipFirstLine) {
         });
     }
 
+    getSecret(lines).terminators = terminators;
+
     return lines;
 };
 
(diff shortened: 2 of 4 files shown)


peet:
On Windows: printing a generated ast will now use `\n` and not `\r\n` (without a explicit `options.lineTerminator = "\r\n"`). This can be fixed by replacing all the `\n` in `lib/printer.js` with `require("os").EOL`. Not sure if worth it?

Also, inserting bits of generated ast in a `\r\n` terminated file will a some point hit the 50% generated lines and switch the whole file to `\n` endings. Food for thought.

fkling:
@benjamn, is this something you would generally support (line break detection)?

alangpierce:
FWIW I did a similar thing in decaffeinate: https://github.com/decaffeinate/decaffeinate/commit/23715d87b2d3e9893c168c75c8241c6014fa531c .

My algorithm was:

1. After reading the file and before doing anything, check the file for the most common type of newline.
2. Convert the code to use `\n` everywhere.
3. All parsing and code generation can assume that `\n` is the only type of newline.
4. Once the new file is generated, convert all newlines back to original newline type detected in step 1.

It's a nice approach because intermediate code never needs to care about newline styles. Although I guess it arguably goes against recast's philosophy of avoiding modifications to the code in places where the AST didn't change.

benjamn:
Yes, I definitely am interested in supporting this, and I'm generally in favor of any cleverness that removes the need for explicit configuration.

It seems like this implementation might call the `guessLineTerminator` function a lot (every time a `Lines` object is sliced, I think). Perhaps that value could be cached, [like we do for `Lines.prototype.guessTabWidth`](https://github.com/peet/recast/blob/0590b079a025256fa7a8c859f785cb42b4099b48/lib/lines.js#L425-L427)?

mjomble:
Although counting all line endings and determining the most often used one seems like a logical way to guess the developer's intent, it might be overkill. I suggest a simpler heuristic:

If a file contains at least one CRLF, use that; in all other cases, use LF.

It's based on the assumption that most projects that use LF will not have a single CRLF in them.

Accidental CRLFs do occur, but then you already had a problem with line endings before recast got involved. Recast would then only amplify the existing problem, which in some situations might even be an improvement since the problem can now be more easily spotted.

Bottom line: as with any other approach, there will be cases where the guess is wrong, but the sheer simplicity of this one (a single substring search) should at least improve performance and maintainability a little bit.

PR #313, opened 2016-09-04
Add spaces around `=` in AssignmentPattern (default parameters)

keyanzhang:
This is a minor stylistic change that adds spaces around `=` in function default parameters. Not sure if you'll take it but it's more aligned with the modern code styles/guidelines. Thanks!


benjamn:
I think I probably cargo-culted the no-spaces style from Python, but I agree that spaces have become the convention in JavaScript. By the way, thank you for simply changing the style to match prevailing conventions, rather than adding a new style option.

keyanzhang:
Thanks Ben!

PR #319, opened 2016-09-17
Removed unnecessary travis ci semvar locks

amilajack:


benjamn:
This was a good idea, but it has happened by other means in the meantime. Sorry for the delay, @amilajack.

PR #340, opened 2016-11-14
A RestElement should not have a trailing comma

keyanzhang:
Seems like the Flow parser and Babylon all recognize `...rest` as a `RestElement`. Which parser uses `.rest` to represent it? :)

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1825,11 +1825,9 @@ function printFunctionParams(path, options, print) {
     if (joined.length > 1 ||
         joined.getLineLength(1) > options.wrapColumn) {
         joined = fromString(",\n").join(printed);
-        if (
-          util.isTrailingCommaEnabled(options, "parameters") &&
-          !fun.rest &&
-          fun.params[fun.params.length - 1].type !== 'RestElement'
-        ) {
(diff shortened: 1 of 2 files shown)


keyanzhang:
friendly ping @benjamn

lib/printer.js
```diff
@@ -1825,7 +1825,11 @@ function printFunctionParams(path, options, print) {
     if (joined.length > 1 ||
         joined.getLineLength(1) > options.wrapColumn) {
         joined = fromString(",\n").join(printed);
-        if (util.isTrailingCommaEnabled(options, "parameters") && !fun.rest) {
+        if (
+          util.isTrailingCommaEnabled(options, "parameters") &&
+          !fun.rest &&
+          fun.params[fun.params.length - 1].type !== 'RestElement'
+        ) {
```
benjamn: Style nit: please line up the conditions vertically after the `if (`:
```js
if (util.isTrailingCommaEnabled(options, "parameters") &&
    !fun.rest &&
    fun.params[fun.params.length - 1].type !== 'RestElement') {
```

keyanzhang: @benjamn thanks, fixed :)

PR #336, opened 2016-10-27
Allow more granular control of trailing commas

mdebbar:
Issue #335

The `trailingComma` option accepts an object that provides more granular control of where to enable/disable trailing commas.

e.g.
```javascript
trailingComma: {
  objects: true, // Enable trailing commas in objects.
  arrays: true, // Enable trailing commas in arrays.
  parameters: false, // Disable trailing commas in function parameters.
}
```

Issue #226, opened 2015-11-02
Allow the pretty-printer to be configured with .eslintrc files


benjamn:
This came up in the discussion over at https://gitter.im/benjamn/recast, thanks to @hzoo.

Though I've known for a long time that the printer isn't very configurable, it seemed like a bad idea to invent a Recast-specific set of configuration options, so I really like the idea of just borrowing the one that `.eslintrc` files already use.

This is potentially a lot of work (adding multiple new cases to each printer rule), but it can be done incrementally, I think.

jamestalmage:
>  This is potentially a lot of work, (adding multiple new cases to each printer rule)

Just to pile on :stuck_out_tongue: , what about `.editorconfig`, and configs to match zero-config linters like `standard`, and `xo`?

I am pretty sure it is possible to coerce all those to `.eslintrc` format (`standard` and `xo` are just `eslint` wrappers, so almost definitely in that case).

benjamn:
I'm happy to pick whichever format can most easily accommodate coerced versions of other formats!

juliankrispel:
Ooooh this would be so neat

skovhus:
What about detecting it based on the existing code? Instead of configuration we can try to detect it by looking at strings in the code (i.e rawValue of string literals).

hzoo:
Yeah that should definetely work for that particular case but it would cover spacing, newlines, etc.

PR #341, opened 2016-11-14
Allow more granular control of spaces inside arrays/objects

keyanzhang:
Let's make everyone happy 😎 . The idea is from #336.

The `spacing` option controls the printing of spaces inside array brackets and braces of object literals, destructuring assignments, and import/export specifiers.

This option could either be:
* Boolean - enable/disable in all contexts (object and arrays).
* Object - enable/disable per context.

Example:
```js
spacing: {
  objects: true,
  arrays: false,
},
```

cc @hzoo

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -75,22 +75,14 @@ var defaults = {
     // }
     trailingComma: false,
 
-    // Controls the printing of spaces inside array brackets and braces of
-    // object literals, destructuring assignments, and import/export specifiers.
-    //
-    // This option could either be:
-    // * Boolean - enable/disable in all contexts (object and arrays).
-    // * Object - enable/disable per context.
-    //
-    // Example:
-    // spacing: {
-    //   objects: true,
-    //   arrays: true,
-    // }
-    spacing: {
-      objects: true,
-      arrays: false,
-    },
+    // Controls the printing of spaces inside array brackets.
+    // See: http://eslint.org/docs/rules/array-bracket-spacing
+    arrayBracketSpacing: false,
+
+    // Controls the printing of spaces inside object literals,
+    // destructuring assignments, and import/export specifiers.
+    // See: http://eslint.org/docs/rules/object-curly-spacing
+    objectCurlySpacing: true,
 
     // If you want parenthesis to wrap single-argument arrow function parameter
     // lists, pass true for this option.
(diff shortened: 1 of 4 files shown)


hzoo:
Awesome looks good!

benjamn:
I like this idea a lot, but I think it might be clearer if we called it `padding`, and if the values could be either `""` or `" "`. Are there any other values that make sense? Maybe `"\n  "`?

keyanzhang:
Yeah, the config name `spacing` is taken from ESLint: http://eslint.org/docs/rules/object-curly-spacing and http://eslint.org/docs/rules/array-bracket-spacing. Happy to change it if `padding` is more clear.

The current setup is only for single-line arrays/objects (although recast doesn't actually support single-line objects at the moment: #208). What would be the benefit of using `''` and `' '`? Do you want it to be a way to specify whether to do single-line or multiline printing (by passing '\n')?

benjamn:
If we're going to go with ESLint conventions (and I think we should—see also https://github.com/benjamn/recast/issues/226), I would very much support two separate options for `array-bracket-spacing` and `object-curly-spacing`.

Sorry, I don't mean to expand the scope of this PR, but I agree that we should stick to ESLint conventions, and maintain my feeling that "spacing" alone is too vague.

keyanzhang:
Thanks Ben, I pushed a new commit. Are you ok with the camel case naming and the boolean value value?

hzoo:
Definetely want to take a look at #226 again

keyanzhang:
Hey @benjamn, is there any chance you can cut a release when you have time? Just wanted to get reactjs/react-codemod#88 fixed -- thanks!

keyanzhang:
@benjamn thanks!

nene:
I wonder why these particular defaults?

Previously Recast didn't add these spaces inside object literals, but now it does. Which made Lebab import/export syntax tests break when run against Recast 0.11.18.

Switched back to old behavior by adding `{objectCurlySpacing: false}` to options. But it took me a while to figure out why my test suddenly started to fail.

keyanzhang:
Sorry @nene! Recast didn't print spaces inside object literals before https://github.com/benjamn/recast/commit/63fbaf97909bbd86a7aaa77bb176c2c124455a84#diff-92f0f9f84108717ce81489fddc9e862aR649 was merged, and my PR added the config options to change it back. I think 63fbaf9 should actually be a breaking change, and when I was choosing the default values I didn't want to break it again.

PR #346, opened 2016-11-29
Print comments when using the generic printer

jlongster:
I noticed that if you use the generic pretty printer it doesn't keep the comments. I'm not sure if this is the right approach, but it works?

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -107,7 +107,7 @@ function Printer(originalOptions) {
 
     // Print the entire AST generically.
     function printGenerically(path) {
-        return genericPrint(path, options, printGenerically);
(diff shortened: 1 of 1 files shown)


lib/printer.js
```diff
@@ -107,7 +107,7 @@ function Printer(originalOptions) {
 
     // Print the entire AST generically.
     function printGenerically(path) {
-        return genericPrint(path, options, printGenerically);
+        return printComments(path, p => genericPrint(path, options, printGenerically));
```
benjamn: Please use a non-arrow function expression here to fix tests in older Node versions. Also the parameter name should be `path` instead of `p`, since the `print` argument gets passed to other code by `printComments`, and we don't always want to be printing the original `path`.

Issue #264, opened 2016-03-03
Failure to properly parenthize short form of arrow function


nene:
When transforming the following AST:

```js
var recast = require('recast');
console.log(recast.print({
    "type": "ExpressionStatement",
    "expression": {
        "type": "AssignmentExpression",
        "operator": "=",
        "left": {
            "type": "Identifier",
            "name": "fn"
        },
        "right": {
            "type": "CallExpression",
            "callee": {
                "type": "MemberExpression",
                "computed": false,
                "object": {
                    "type": "ArrowFunctionExpression",
                    "id": null,
                    "params": [],
                    "defaults": [],
                    "body": {
                        "type": "Identifier",
                        "name": "a"
                    },
                    "generator": false,
                    "expression": true
                },
                "property": {
                    "type": "Identifier",
                    "name": "bind"
                }
            },
            "arguments": [
                {
                    "type": "ThisExpression"
                }
            ]
        }
    }
}).code);
```

expected output:

```js
fn = (() => a).bind(this);
```

actual output:

```js
fn = () => a.bind(this);
```

typeetfunc:
I find similar bug:

```Javascript
const recast = require('recast');
const b = recast.types.builders;
const ast = recast.parse(`props => {}`)
var propsParam = b.identifier('props')
propsParam.typeAnnotation = b.typeAnnotation(b.genericTypeAnnotation(b.identifier('Props'), null))
var arrowFunc = b.arrowFunctionExpression(
        [propsParam],
        b.blockStatement([])
)

ast.program.body[0].expression = arrowFunc
var ast1 = b.file(b.program(
    [b.expressionStatement(arrowFunc)]
))
console.log(recast.print(ast).code) // props: Props => {}
console.log(recast.print(ast1).code) // (props: Props) => {};
```

mdebbar:
@benjamn I made a PR yesterday to fix this issue. PR also includes a test to make sure the behavior is correct, and to prevent future regressions. Please take a look whenever you have some time.

mdebbar:
@benjamn @nene this issue can be closed now. It's fixed in v0.11.20

PR #349, opened 2016-12-10
Surround arrow functions with parenthesis when binding

mdebbar:
Fixes the critical, long standing issue: #264 
Also fixes https://github.com/lebab/lebab/issues/105

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -326,10 +326,12 @@ FPp.needsParens = function(assumeExpressionContext) {
         }
 
     case "ArrowFunctionExpression":
-        if(parent.type === 'CallExpression' && 
-           name === 'callee') {
+        if(n.CallExpression.check(parent) && name === 'callee') {
             return true;
-        };
+        }
+        if(n.MemberExpression.check(parent) && name === 'object') {
+            return true;
+        }
 
         return isBinary(parent);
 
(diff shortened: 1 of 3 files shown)


benjamn:
Though this implementation appears to work (and I thank you for submitting it), the right place to fix problems with parentheses is [`FastPath.prototype.needsParens`](https://github.com/benjamn/recast/blob/cbdb59b4fa081f1dfd7f7477bb4e98040d5e5178/lib/fast-path.js#L179), specifically [this case](https://github.com/benjamn/recast/blob/cbdb59b4fa081f1dfd7f7477bb4e98040d5e5178/lib/fast-path.js#L328).

Adding logic to `needsParens` ensures that we get the parentheses right not just for pretty-printed code, but for any reprinted code, e.g. if an arrow function gets moved from somewhere in the original AST where it didn't need parentheses to a position where it needs parentheses.

Can you move the logic to `needsParens` (and keep the tests)?

mdebbar:
Updated, sir! 🙂
Thanks for suggesting the right way to do it.

shajith:
@benjamn: Could this be merged/tagged? Looking for a fix for the referenced lebab/lebab#105

Thanks!

shajith:
@benjamn: Thank you!

PR #352, opened 2016-12-17
fixes #351 - parenthesis around object expression and conditional expression

### LINKED ISSUE #351: missing parentheses in printed code causes syntax error
```javascript
recast.print({
    type: "AssignmentExpression",
    left: {
        type: "MemberExpression",
        computed: false,
        object: {type: "ObjectExpression", properties: []},
        property: {type: "Identifier", name: "x"}
    },
    operator: "=",
    right: {type: "Literal", value: 42}
}).code // "{}.x = 42"
```

I checked that parsing the expected result with correct parentheses
```javascript
recast.parse("({}).x=42") 
```
produces the same AST (which is printed with parentheses), the only difference is the existence of "original" and "loc" properties.
It seems like the printer thinks it is not necessary to print parentheses, but when the parsed expression has parentheses, it prints them.

I discovered a similar problem:

```javascript
var callee = recast.parse('typeof a==="function"?a:b').program.body[0].expression;
recast.print({type:"NewExpression", callee:callee, arguments:[]); // 'new typeof a==="function"?a:b()' 
```

In this case, it seems like the printer thinks it is neccessary to print parentheses (when i don't use a parsed expression object) but because the parsed expression hasn't, it does not print any.

zoltan-mihalyi:


zoltan-mihalyi:
Is this the correct solution?

zoltan-mihalyi:
Can you merge it?

test/printer.js
```diff
@@ -1605,4 +1605,41 @@ describe("printer", function() {
...
+
+        var printer = new Printer({
+            arrowParensAlways: true
+        });
+        var pretty = printer.printGenerically(ast).code;
+        assert.strictEqual(pretty, code);
+    });
+
+    it("adds parenthesis around conditional", function() {
+        var code = 'new ((typeof a ? b : c))();';
```
benjamn: Why the extra parentheses here? Is this actually desirable, or are we erring on the side of safety (and this is just an edge case)?

zoltan-mihalyi: There is bug responsible for the duplicated parentheses in my code: #366
After it is fixed, the extra parentheses should be removed.

benjamn:
Thanks @zoltan-mihalyi! Sorry for the wait.

PR #353, opened 2016-12-24
`allowBreak` adds an extra newline.

pksjce:
I was facing this issue in [this example](http://astexplorer.net/#/iBjDDgDrsg) , where if an `ObjectExpression` has a property of type `ObjectExpression` then the printer adds an extra new line on the property.

Added a test case for the printer where you can see it fails.
Added a fix for it where if I don't set `allowBreak`, no other test fails and my test passes.
The variables `multiLine`, `allowBreak`, `oneLine` are not clear to me. How can I fix this properly?

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -637,7 +637,10 @@ function genericPrintNoParens(path, options, print) {
                     // Add an extra line break if the previous object property
                     // had a multi-line value.
                     parts.push(separator + (multiLine ? "\n\n" : "\n"));
-                    allowBreak = !multiLine;
+                    /*
+                        Commenting this passes my test case and breaks nothing else.
+                        allowBreak = !multiLine;
+                    */
                 } else if (len !== 1 && isTypeAnnotation) {
                     parts.push(separator);
                 } else if (!oneLine && util.isTrailingCommaEnabled(options, "objects")) {
--- a/test/printer.js
+++ b/test/printer.js
@@ -364,6 +364,32 @@ describe("printer", function() {
         ].join(eol));
     });
 
+    it("MultiLineVarWithObjPrinting", function() {
+        var printer = new Printer({ tabWidth: 2 });
+        var varDecl = b.variableDeclaration("var", [
+            b.variableDeclarator(b.identifier("x"), null),
+            b.variableDeclarator(
+                b.identifier("y"),
+                b.objectExpression([
+                    b.property("init", b.identifier("why"), b.literal("not")),
+                    b.property("init", b.identifier("options"), b.objectExpression([b.property("init", b.identifier("what"), b.literal("else"))]))
+                ])
+            ),
+            b.variableDeclarator(b.identifier("z"), null)
+        ]);
+
+        assert.strictEqual(printer.print(b.program([varDecl])).code, [
+            "var x,",
+            "    y = {",
+            "      why: \"not\",",
+            "      options: {",
+            "        what: \"else\"",
+            "      }",
+            "    },",
+            "    z;"
+        ].join(eol));
+    });
+
     it("ForLoopPrinting", function() {
         var printer = new Printer({ tabWidth: 2 });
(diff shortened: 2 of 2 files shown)


pksjce:
@benjamn - Can you help me out here? Due to this isssue I'm forced to use a branch of recast with my fix :(

alangpierce:
Hi @pksjce, I'm not @benjamn, but I've been working with this project a little recently and thought I'd look into this and maybe help out.

Looks like this really old commit is where `allowBreak` was introduced:
https://github.com/benjamn/recast/commit/4d62420272523286ed76319f7e1796dcbf76c795#diff-92f0f9f84108717ce81489fddc9e862aR144

It's a little easier to see what's going on without all of the changes since then, and the "Similar to the logic in Syntax.BlockStatement" actually references another `allowBreak` variable further down in the file that has a little more explanation.

From what I can tell, the code is intentionally putting a blank line above and below any multiline object properties. The `allowBreak` boolean communicates to the next loop iteration to avoid adding two blank lines if there are two multiline properties in a row.

So at least from my reading of things, the code is working as expected, and the extra newline is just a style preference in the recast printer. You may be able to argue for changing the default style (I certainly don't put blank lines around multiline properties), but I think in general recast tries to just generate "something reasonable" rather than exposing a ton of style options.

Maybe an alternative solution would be to run a follow-up reformatter for your particular style. Often `eslint --fix` is good enough to fix the resulting code, but unfortunately I don't see an eslint option for this case. But you might explore other formatters like [esformatter](https://github.com/millermedeiros/esformatter), if you haven't already.

Another option is to just write your own script that removes blank lines between adjacent object properties and run it afterward. You can't use recast for that, so I would use jscodeshift and [magic-string](https://github.com/Rich-Harris/magic-string) (since jscodeshift gives you easy access to the AST and doesn't force you to use the recast printer to produce the output). Basically you'd search the AST for object properties, then remove blank lines using `magicString.remove`, then return `magicString.toString()`.

pksjce:
Thank you @alangpierce for the explanation.

However I still would argue that I still don't see any value in adding extra newlines. It would be great to change this default styling as multiple people have faced this problem and it is clearly an unexpected result.

Also, adhering to recast's codebase, I was able to create a failing [test case](https://github.com/benjamn/recast/pull/353/files#diff-0fec46963312f671a23f0db5db5b5ad0R367).  I feel its hacky to use another tool and especially `eslint --fix` for this purpose.

benjamn:
Thanks for asking about this issue, @pksjce, and thanks for investigating and explaining the background, @alangpierce!

As @alangpierce correctly surmises, putting newlines around multiline statements and properties is a deliberate style choice that I made in the early history of this project, and it's a style that I continue to prefer. One day, when I have lots of free time, the pretty-printer may become the fully configurable tool of everyone's dreams. In the meantime, it's a somewhat opinionated tool that reflects my personal style. See https://github.com/benjamn/recast/issues/226 for more on that long-term vision.

Rather than convincing you that my style is better, or asking you to persuade me otherwise, I hope I can persuade you that this is not a binary choice. We actually have several options:

1. Preserve the status quo (continue putting newlines around multi-line properties).
2. Abandon the status quo (stop putting newlines around multi-line properties, as in this PR).
3. Introduce a new printing option that allows the style choice to be configured.
4. Do a better job preserving existing style without manual intervention.

Among these options, my ordering of preferences is almost always 4 > 1 > 3 > 2, though sometimes 4 is too much work, so 1 prevails.

Why is 3 not more appealing? Accumulating one-off printer options over time makes the API more and more Recast-specific, which increases the slope of the learning curve without ever reaching the goal of total flexibility. Again, see #226 for a more principled approach to printer configuration.

In this case, I believe there is a reasonable way to achieve the 4th option. Please see [this comment](https://github.com/benjamn/recast/issues/370#issuecomment-278970141) for a recent discussion of the same problem, together with a potential solution.

PR #354, opened 2016-12-29
Add sourceFileName passthrough to File node

BrendanAnnable:
This PR passes the `sourceFileName` through to the `File` node in `ast-types`.

It is dependant on `name` being added as an optional attribute to the `File` node. It thankfully is backwards compatible (in that it won't break anything) as it appears `ast-types` ignores any extra params.

The `ast-types` PR can be found at https://github.com/benjamn/ast-types/pull/200. More details of its use case there.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -63,7 +63,7 @@ exports.parse = function parse(source, options) {
     // comments, wrap the original Program node with a File node.
     var file = program;
     if (file.type === "Program") {
-        var file = b.file(program);
+        var file = b.file(program, options.sourceFileName);
         file.loc = {
             lines: lines,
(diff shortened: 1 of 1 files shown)


lib/parser.js
```diff
@@ -63,7 +63,7 @@ exports.parse = function parse(source, options) {
     // comments, wrap the original Program node with a File node.
     var file = program;
     if (file.type === "Program") {
-        var file = b.file(program);
+        var file = b.file(program, options.sourceFileName);
```
benjamn: Maybe coerce this to `null` if falsy?
```js
b.file(program, options.sourceFileName || null);
```

PR #358, opened 2017-01-09
Add option to avoid copying the AST

jlongster:
In my project, I'm using recast to parse a large file. The file is 2000 lines long. Running my project takes ~970ms. I profiled it under Chrome:

screen shot 2017-01-09 at 2 47 46 pm

You can see the `TCp.copy` function is taking a large chunk of the time. I looked into it and it's because recast copies the entire ast before returning it. Turning this off, my project runs in ~660ms, about 2/3 the time. Copying a big AST has a big performance penalty so I think it's worth making this an option, and possibly even defaulting to `false`.

Should the default be `false`? If so many projects using recast would see a 30% perf bump.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -92,6 +92,11 @@ var defaults = {
     // The use of commas is in line with the more popular style and matches
     // how objects are defined in JS, making it a bit more natural to write.
     flowObjectCommas: true,
+
+    // By default, recast copies the entire AST so you can compare
+    // changes. This has a significant performance cost, however, so
+    // you can turn it off it you don't need it.
+    copyAST: true,
 }, hasOwn = defaults.hasOwnProperty;
 
 // Copy options and fill in default values.
@@ -123,5 +128,6 @@ exports.normalize = function(options) {
         objectCurlySpacing: get("objectCurlySpacing"),
         arrowParensAlways: get("arrowParensAlways"),
         flowObjectCommas: get("flowObjectCommas"),
+        copyAST: !!get("copyAST")
     };
 };
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -84,8 +84,12 @@ exports.parse = function parse(source, options) {
     );
 
     // Return a copy of the original AST so that any changes made may be
-    // compared to the original.
-    return new TreeCopier(lines).copy(file);
+  // compared to the original.
+    if(options.copyAST) {
+      return new TreeCopier(lines).copy(file);
+    } else {
+      return file;
+    }
 };
 
 function TreeCopier(lines) {


benjamn:
I disagree strongly with this idea. Copying the AST is necessary so that the modified AST can be compared to the original AST, so that Recast can do its primary job: conservatively reprinting the new AST. If you take away that functionality, there's not much point in using Recast. That said, I'm happy to consider any ideas for speeding up the process of creating the shadow AST, or even new ideas that somehow create it lazily, but this PR just doesn't make sense.

benjamn:
More generally, while parsing/printing speed is important, and has improved dramatically over time (e.g. thanks to the `FastPath` abstraction, which is much less convenient than the `NodePath` abstraction it replaced), the goal of Recast is to minimize unnecessary changes to the original code, which allows it to produce compact, human-reviewable diffs. Sacrificing the quality of reprinted code for performance is not a tradeoff that makes sense for this project, because the speed of running an automatic codemod is rarely the bottleneck when you're refactoring.

jlongster:
> If you take away that functionality, there's not much point in using Recast

Oh, that's true. I hadn't realized that I'm not really using much of recast anymore, because I'm using the babylon parser anyway. I can probably use it directly. It makes sense that recast would do it as the core functionality depends on it.

I'm working on a project that is the opposite of the recast printer: it strips away all custom style and pretty-prints your code so that's deterministically formatted and you can enforce that across your project. I liked recast because it handles a lot of edge cases already, so that's why I'm using it. But my printer doesn't need the original nodes at all, so I can use babylon + my printer.

I do lose some functionality. The way you attach comments, fix a bunch faulty locations, and things like that. I may have to copy that code into my printer and just use it there. Previously I was getting this for free by going through recast, but not sure I want to pay this perf penalty when I don't need it.

benjamn:
Yeah, you should be able to pass any AST to `recast.print`, even if it didn't come from `recast.parse`.

jlongster:
Well, my project is a new printer, so now it's just babylon -> my printer :) But I want the `attachComments` functionality from recast so need to figure out how to borrow that code.

benjamn:
@cspotcode You can totally write your transformers that way with Recast. I just wanted this tool to work equally well regardless of how you decide to implement your transforms.

Incidentally, I wonder if the shadow AST technique could be avoided entirely by wrapping the original AST in a `Proxy`-based membrane, so that you could keep track of what parts of the AST were modified without making full copy. When I first started working on Recast, the `Proxy` API did not exist, and it still isn't safe to use in all browsers, but Recast typically runs in Node, not Internet Explorer. If anyone is interested in exploring this approach, I would love to collaborate, and I have this excellent npm package name that we can use: https://www.npmjs.com/package/brane

cspotcode:
You might hit performance issues where traversing the AST requires
essentially constructing a proxy for every node of the tree.  And *all*
object access would happen via proxies; I'm not sure how performance
compares to native object access.  But if the performance proves to be
improved...

You might be able to write a `set` & `delete` trap that shallow-clones the
node onto itself:  `node[PristineClone] = {...node}` when it's modified for
the first time.

On Oct 17, 2018 6:30 PM, "Ben Newman" <notifications@github.com> wrote:

@cspotcode <https://github.com/cspotcode> You can totally write your
transformers that way with Recast. I just wanted this tool to work equally
well regardless of how you decide to implement your transforms.

Incidentally, I wonder if the shadow AST technique could be avoided
entirely by wrapping the original AST in a Proxy-based membrane, so that
you could keep track of what parts of the AST were modified without making
full copy. When I first started working on Recast, the Proxy API did not
exist, and it still isn't safe to use in all browsers, but Recast typically
runs in Node, not Internet Explorer. If anyone is interested in exploring
this approach, I would love to collaborate, and I have this excellent npm
package name that we can use: https://www.npmjs.com/package/brane

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/benjamn/recast/pull/358#issuecomment-430812116>, or mute
the thread

cspotcode:
> @cspotcode You can totally write your transformers that way with Recast. I just wanted this tool to work equally well regardless of how you decide to implement your transforms.

My point was, you won't get any performance benefit from writing transformations in that style *unless* you're somehow able to disable recast's eager AST cloning.

PR #363, opened 2017-01-21
Fix precedence of exponentiation operator

nene:
I implemented it by myself in PR #285

But somehow ended up adding it to the same precedence category
as `*`, `/`, `%`.  Turns out `**` is in a precendence category of its own,
above the `*`, `/`, `%`.

The spec is somewhat hard to digest:
- https://www.ecma-international.org/ecma-262/7.0/#sec-multiplicative-operators

A more firendly reference is provided by MDN:
- https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -380,7 +380,8 @@ var PRECEDENCE = {};
  ["<", ">", "<=", ">=", "in", "instanceof"],
  [">>", "<<", ">>>"],
  ["+", "-"],
- ["*", "/", "%", "**"]
+ ["*", "/", "%"],
+ ["**"]
 ].forEach(function(tier, i) {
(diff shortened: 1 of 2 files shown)


nene:
Actually this solution is not fully correct either because `**` is right-associative unlike e.g. multiplication that's left-associative.

So, when you create an AST like: `((x ** y) ** z)`

Recast will print it out as: `x ** y ** z` which means `(x ** (y ** z))`

It seems Recast doesn't really have a system to deal with associativity. Rather the operators with different associativity have so far have conveniently had different node types. I guess all the BinaryExpression nodes so far have been left-associative, but `**` operator is an exception.

PR #364, opened 2017-01-22
Avoid an ASI bug by sometimes wrapping return expressions in parens

### LINKED ISSUE #362: Adding a comment to a return expression causes incorrect output
If I start with this code:

```js
function f() {
  return 3;
}
```

and add a `// foo` comment on the `3` node, recast prints this result:

```js
function f() {
  return //foo
  3;
}
```

This changes the behavior, since the ASI rules cause the function to return `undefined`.

Here's an astexplorer test case:
https://astexplorer.net/#/m2KsWvHL0l

I'm taking a look at this and hoping to be able to come up with a PR, but any thoughts or pointers would be helpful.

alangpierce:
Fixes #362

If the argument to a return statement starts with a comment-only line, then we
now wrap it in parens to avoid the `return` getting parsed as its own line:

```js
function f() {
  return (
    // Foo
    3
  );
}
```

This problem can be introduced by adding a leading comment to any leftmost child
of the return argument, so we need to be more defensive about when we reformat
`return` statements. For now, we reprint the `ReturnStatement` node (but leave
unchanged children intact) when any child had changes.

### DIFF
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -458,7 +458,9 @@ Lp.isFirstLineOnlyComment = function () {
         return false;
     }
     var firstLineInfo = secret.infos[0],
-        firstLine = firstLineInfo.line.trim();
+        sliceStart = firstLineInfo.sliceStart,
+        sliceEnd = firstLineInfo.sliceEnd,
+        firstLine = firstLineInfo.line.slice(sliceStart, sliceEnd).trim();
     return firstLine.length === 0 ||
         firstLine.slice(0, 2) === "//" ||
         firstLine.slice(0, 2) === "/*";
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -448,6 +448,22 @@ Lp.guessTabWidth = function() {
     return secret.cachedTabWidth = result;
 };
 
+// Determine if the list of lines has a first line that's empty of non-comment
+// tokens. If this is the case, the code may need to be wrapped in parens to
+// avoid ASI issues. Calling code should be more defensive if we return true,
+// so false positives are ok.
+Lp.isFirstLineOnlyComment = function () {
+    var secret = getSecret(this);
+    if (secret.infos.length === 0) {
+        return false;
+    }
+    var firstLineInfo = secret.infos[0],
+        firstLine = firstLineInfo.line.trim();
+    return firstLine.length === 0 ||
+        firstLine.slice(0, 2) === "//" ||
+        firstLine.slice(0, 2) === "/*";
+};
+
 Lp.isOnlyWhitespace = function() {
     return isOnlyWhitespace(this.toString());
 };
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -4,6 +4,7 @@ var types = require("./types");
 var getFieldValue = types.getFieldValue;
 var Printable = types.namedTypes.Printable;
 var Expression = types.namedTypes.Expression;
+var ReturnStatement = types.namedTypes.ReturnStatement;
 var SourceLocation = types.namedTypes.SourceLocation;
 var util = require("./util");
 var comparePos = util.comparePos;
@@ -518,6 +519,13 @@ function findChildReprints(newPath, oldPath, reprints) {
         if (!canReprint) {
             return false;
         }
+    }
+
+    // Return statements might end up running into ASI issues due to comments
+    // inserted deep within the tree, so reprint them if anything changed
+    // within them.
+    if (ReturnStatement.check(newPath.getNode()) && reprints.length > 0) {
+        return false;
     }
 
     return true;
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -509,6 +509,8 @@ function findChildReprints(newPath, oldPath, reprints) {
     // Don't bother traversing .loc objects looking for reprintable nodes.
     delete keys.loc;
 
+    var originalReprintCount = reprints.length;
+
     for (var k in keys) {
         newPath.stack.push(k, types.getFieldValue(newNode, k));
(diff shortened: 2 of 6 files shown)


alangpierce:
Thoughts on this? :-)

I have this patch locally on my computer to avoid a bug when converting CoffeeScript files without JSX to JavaScript with JSX, and it would be nice if others on my team could run the full transform without having to use local forks of recast and jscodeshift. I also have a fix for [create-element-to-jsx](https://github.com/reactjs/react-codemod/blob/master/transforms/create-element-to-jsx.js) that's blocked on this (it improves comment positioning, but would introduce bugs without this fix applied).

lib/lines.js
```diff
@@ -448,6 +448,22 @@ Lp.guessTabWidth = function() {
...
+// tokens. If this is the case, the code may need to be wrapped in parens to
+// avoid ASI issues. Calling code should be more defensive if we return true,
+// so false positives are ok.
+Lp.isFirstLineOnlyComment = function () {
+    var secret = getSecret(this);
+    if (secret.infos.length === 0) {
+        return false;
+    }
+    var firstLineInfo = secret.infos[0],
+        firstLine = firstLineInfo.line.trim();
```
benjamn: This method has some (admittedly non-obvious) problems. For example,
* The `secret.infos[i].line` string may have been inherited from another `Lines` object, so it needs to be sliced from `secret.infos[i].sliceStart` to `secret.infos[i].sliceEnd`, then trimmed.
* What if `firstLine.startsWith("/*")` but then later has a `*/` followed by non-comment tokens?

That said, I think we can do without the method, so no need to fix it.

lib/printer.js
```diff
@@ -568,9 +568,11 @@ function genericPrintNoParens(path, options, print) {
 
         if (n.argument) {
             var argLines = path.call(print, "argument");
-            if (argLines.length > 1 &&
-                namedTypes.JSXElement &&
-                namedTypes.JSXElement.check(n.argument)) {
+            if (argLines.isFirstLineOnlyComment() ||
```
benjamn: Could we just check if
```js
n.comments && n.comments.some(function (comment) {
  return comment.leading && comment.type.startsWith("Line");
});
```

benjamn: This might give some false positives, but the safety seems worth it.

benjamn:
Thanks for working on this, by the way!

lib/printer.js
alangpierce: That actually misses some cases. I think you'd need `n.argument.comments`, not `n.comments`, but even then it only gets comments on the top-level node. If, for example, you have a binary operator and a leading comment on the LHS, then the expression would still need to be wrapped in parens. I originally was thinking of doing an AST-based approach like this, but I think it requires repeatedly traversing the leftmost child and checking for any leading comments on that path, which seems non-trivial. Just doing string matching on the source code seems simpler and more robust in a certain sense (since the source code is what the parser is going to see).

I didn't have a test for the LHS binary operator case, but I'll add one.

lib/lines.js
alangpierce: > The `secret.infos[i].line` string may have been inherited from another `Lines` object, so it needs to be sliced from `secret.infos[i].sliceStart` to `secret.infos[i].sliceEnd`, then trimmed.

Thanks, I'll update the PR to fix that.

> What if `firstLine.startsWith("/*")` but then later has a `*/` followed by non-comment tokens?

Then it would be a false positive and wrapped in parens when it technically doesn't need to be, which is still correct. (See the comment on the function for a little more of the rationale.) It doesn't seem like there's a great way to detect this case with just string matching (while also returning true for cases like `/* a */ /*`), and doing real lexing seems like needless complexity, but happy to hear suggestions on ways to improve the check.

lib/printer.js
benjamn: Got it. I'm happy to keep iterating with you on `Lines#isFirstLineOnlyComment`, in that case!

alangpierce:
@benjamn Thanks for taking a look! I updated the PR with a test to demonstrate that your `n.comments.some` suggestion actually isn't sufficient, and I updated the code to use `sliceStart`/`sliceEnd`.

lib/lines.js
benjamn: > Then it would be a false positive and wrapped in parens when it technically doesn't need to be, which is still correct.

Agreed.

lib/patcher.js
```diff
@@ -520,5 +521,12 @@ function findChildReprints(newPath, oldPath, reprints) {
         }
     }
 
+    // Return statements might end up running into ASI issues due to comments
+    // inserted deep within the tree, so reprint them if anything changed
+    // within them.
+    if (ReturnStatement.check(newPath.getNode()) && reprints.length > 0) {
```
benjamn: Although this is safe as written, let's save `reprints.length` before the loop and check if it has increased here, so that we only `return false` if the argument changed (there may be other reprints from elsewhere in the tree).

alangpierce: Ok, I just pushed a commit that does that.

benjamn:
Thanks again for working on this! Just a heads up, in case it matters to you: I'm planning to rename `Lines#isFirstLineOnlyComment` to `Lines#startsWithComment`, since I think that indicates there could be other tokens on the line.

benjamn:
Published to npm as [v0.11.22](https://github.com/benjamn/recast/releases/tag/v0.11.22).

alangpierce:
@benjamn thanks for reviewing and publishing! Sure, the rename makes sense, although keep in mind that it should (and does) return true if the first line is completely empty (since that's another case where a ASI would change the AST).

Issue #377, opened 2017-03-12
For loop printed with extra semicolon when initialized with `let` and body replaced with block


outoftime:
Probably easiest to start with a reduction:

```js
const recast = require('recast');
const b = recast.types.builders;

const ast = recast.parse(`
  for (let i = 0; i < 1; i++)
    console.log(i);
`);
const loop = ast.program.body[0];

loop.body = b.blockStatement([]);

console.log(recast.print(ast).code);
```

The output is:

```js
for (let i = 0;; i < 1; i++)
  {}
```

Note the extra semicolon after the loop initialization statement.

From playing around, in order to see this bug, you need to:

* Initialize the loop with a `let` (works fine with a `var`)
* Have the original loop contain a single-statement, non-block body
* Replace the body with a block statement (doesn’t matter what’s in the block)

Observed with recast 0.11.22

benjamn:
This appears to be a parser bug, since `loop.init.loc.{start,end}` wrongly includes the `;` character from the original code. In other words, `loop.init.loc.end.column` is one character too large.

Fortunately, parser bugs that lead to faulty `.loc.{start,end}` information can usually be fixed in [`util.fixFaultyLocations`](https://github.com/benjamn/recast/blob/412aaf89d3993a58b35c8f255a1c6751540064d9/lib/util.js#L147).

PR #379, opened 2017-03-13
[test] Ensure both parentheses in a parenthesised unary expression

abernix:
This test-only (no fix included, sorry!) is mainly submitted for consideration/comment.

While parentheses are properly maintained when coming from `babylon`, the trailing parenthesis is stripped when printing an AST originally from `flow-parser` (which I added to `devDependencies`) when it has been mutated.

In this case (as demonstrated in the test), this code when transformed:

    function pem() { return !(broke && welsh); }

Is `print`-ed again as:

    function pem() { return !(broke && welsh; }
    //                                      `--- ) omitted.

I'm happy to address an actual fix but didn't familiarize myself with the rest of the codebase much.  `FPp.needsParens`, maybe, but seems like this was more a matter of a parenthesis getting removed when it should have been preserved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "babel-preset-es2015": "^6.22.0",
     "babylon": "~6.15.0",
     "esprima-fb": "^15001.1001.0-dev-harmony-fb",
+    "flow-parser": "^0.40.0",
     "mocha": "~3.1.2",
     "reify": "^0.4.16"
   },
--- a/test/printer.js
+++ b/test/printer.js
@@ -1015,6 +1015,37 @@ describe("printer", function() {
     assert.strictEqual(pretty, code);
   });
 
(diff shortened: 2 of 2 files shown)


benjamn:
I suspect this bug is analogous to https://github.com/benjamn/recast/issues/377, which I recently [fixed](https://github.com/benjamn/recast/commit/69475b749da6db08ddde678a1550ba7d7da168f8) by adding another case to `util.fixFaultyLocations`.

When we reuse the original source code, it's important for the `node.loc.{start,end}` information to be correct, because that's what we use to slice the right code from the original string. Unfortunately, parsers tend to get `.loc` information slightly wrong from time to time, probably because not many clients actually rely on that metadata.

The good news is that Recast has been able to fix all such problems unilaterally, using the aforementioned `util.fixFaultyLocations` function.

PR #393, opened 2017-04-04
Add parens around Babel's numeric literals.

loganfsmyth:
From https://github.com/babel/babel/issues/5579

    b.memberExpression({type: "Literal", value: 1}, b.identifier('foo'))

outputs

    (1).foo

but

    b.memberExpression({type: "NumericLiteral", value: 1}, b.identifier('foo'))

outputs

    1.foo

which is a syntax error.

Not sure where tests for this might belong, let me know if you want me to add some.


benjamn:
Thanks @loganfsmyth!

Issue #315, opened 2016-09-07
useTabs option works only partially


nene:
When running the following code:

```js
var recast = require('recast');

var ast = recast.parse(
  'while (true) {\n' +
  '\tx;\n' +
  '}'
, {useTabs: true});

// Add one line inside WHILE loop
ast.program.body[0].body.body.push({
  'type': 'ExpressionStatement',
  'expression': {
    'type': 'Identifier',
    'name': 'y'
  }
});

console.log(recast.print(ast, {useTabs: true}).code.replace(/\t/g, "<TAB>"));
```

I'd expect to get:

```
while (true) {
<TAB>x;
<TAB>y;
}
```

But instead Recast gives me:

```
while (true) {
    x;
<TAB>y;
}
```

NOTE: There are no docs to tell me whether I should pass `useTabs` option to `recast.parse()` or `recast.print()`, it seems that it's only needed for the latter, but just in case I'm passing it to both.

PR #404, opened 2017-05-10
Respect useTabs option when indenting an opening JSX element

jsnajdr:
Steps to reproduce:
Run this node script:
```
const recast = require('.');

// parse some code into AST
let ast = recast.parse([
	"function x(a, b) {",
	"\treturn (",
	"\t\t<div>",
	"\t\t\t{ b }",
	"\t\t</div>",
	"\t);",
	"}"
].join("\n"));

// change the variable identifier in JSX to trigger reprint
ast.program.body[0].body.body[0].argument.children[1].expression.name = 'a';

// print the output with useTabs=true
let output = recast.print(ast, { useTabs: true }).code;

// replace tabs with '~' and spaces with '_'
output = output.replace(/ /g, "_").replace(/\t/g, "~");

// print the output code with visible whitespace
console.log(output);
```

Expected result:
The output is indented with tabs:
```
function_x(a,_b)_{
~return_(
~~<div>
~~~{_a_}
~~</div>
~);
}
```

Actual result:
There are spaces before the JSX opening element:
```
function_x(a,_b)_{
~return_(
~____<div>
~~~{_a_}
~~</div>
~);
}
```

The cause of the bug is:
- in return statement, a JSX element argument gets some [extra indentation](https://github.com/benjamn/recast/blob/master/lib/printer.js#L581)
- the `Lines.prototype.join` method [uses spaces to create the indent](https://github.com/benjamn/recast/blob/master/lib/lines.js#L823) and doesn't respect the `useTabs` option

The fix is to use a smarter algorithm from the `Lines.prototype.sliceString` function.

However, this requires passing a new parameter to the `join` method and recursively add `options` to many many calls. I'm wondering if there is a better solution for this.

### DIFF
--- a/lib/lines.js
+++ b/lib/lines.js
@@ -751,7 +751,6 @@ Lp.sliceString = function(start, end, options) {
 
     var infos = getSecret(this).infos;
     var parts = [];
-    var tabWidth = options.tabWidth;
 
     for (var line = start.line; line <= end.line; ++line) {
         var info = infos[line - 1];
@@ -777,23 +776,7 @@ Lp.sliceString = function(start, end, options) {
             continue;
         }
 
-        var tabs = 0;
-        var spaces = indent;
-
-        if (options.useTabs) {
-            tabs = Math.floor(indent / tabWidth);
-            spaces -= tabs * tabWidth;
-        }
-
-        var result = "";
-
-        if (tabs > 0) {
-            result += new Array(tabs + 1).join("\t");
-        }
-
-        if (spaces > 0) {
-            result += new Array(spaces + 1).join(" ");
-        }
+		var result = indentString(indent, options);
 
         result += info.line.slice(info.sliceStart, info.sliceEnd);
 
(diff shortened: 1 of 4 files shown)


lib/lines.js
```diff
@@ -807,7 +790,29 @@ Lp.isEmpty = function() {
...
+
+	if (tabs > 0) {
+		result += new Array(tabs + 1).join("\t");
+	}
+
+	if (spaces > 0) {
+		result += new Array(spaces + 1).join(" ");
+	}
+
+	return result;
```
benjamn: Can you stick to four spaces per tab when adding new implementation code, please? I realize you're a fan of tabs (nothing wrong with that), but consistency is important here.

jsnajdr:
> Can you stick to four spaces per tab when adding new implementation code, please? I realize you're a fan of tabs (nothing wrong with that), but consistency is important here.

Oh, sorry for submitting broken commits like this :( I pushed a tabs-to-spaces change, and also changed the `indentString` function to avoid param destructuring -- the tests were failing on older versions of Node.js that don't have good enough ES6 support.

lib/lines.js
Hexcede: > Can you stick to four spaces per tab when adding new implementation code, please? I realize you're a fan of tabs (nothing wrong with that), but consistency is important here.

I apologize in advance, but I really can't help emphasizing the enormous amount of irony in the statement "consistency is important here" in reference to indents in a PR that is targeting a design flaw that produces inconsistent indents.

Could issue #315 as well as this PR possibly be addressed since this project appears maintained? These issues have been around for six years now.

PR #410, opened 2017-06-04
Don't wrap IIFEs in parens when using babylon

xixixao:
This is not the right fix, but i couldn't figure out how to fix it properly. Babylon now tags nodes with an object:

```
extra: {
  parenthesized: true,
  parenStart: 0
}
```
I think Recast confuses it as a Node and then tries to reprint instead of using original node, causing iifes to be wrapped in parens.

### DIFF
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -525,7 +525,7 @@ function findChildReprints(newPath, oldPath, reprints) {
   var originalReprintCount = reprints.length;
 
   for (var k in keys) {
-    if (k.charAt(0) === "_") {
+    if (k.charAt(0) === "_" || k === 'extra') {
       // Ignore "private" AST properties added by e.g. Babel plugins and
       // parsers like Babylon.
       continue;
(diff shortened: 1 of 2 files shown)


benjamn:
Thanks for reporting this. I need to step through the code to see what's happening with that `extra` field, since at first glance I don't know why it would be considered changed (thus requiring reprinting).

lib/patcher.js
```diff
@@ -498,10 +499,12 @@ function findChildReprints(newPath, oldPath, reprints) {
   // path.needsParens() && !hasParens(oldNode) check below, the absence of
   // a closing parenthesis after the FunctionExpression would trigger
   // pretty-printing unnecessarily.
-  if (!newPath.canBeFirstInStatement() &&
+  if (Node.check(newNode) &&
+      !newPath.canBeFirstInStatement() &&
       newPath.firstInStatement() &&
-      !hasOpeningParen(oldPath))
+      !hasOpeningParen(oldPath)) {
     return false;
```
benjamn: @xixixao Your test case was triggering this `return false` when `newNode` is the `extra` object, though this logic only applies to `Node` objects.

benjamn:
@xixixao Can you test that this works with `recast@0.12.5` (just released)?

xixixao:
@benjamn confirmed your fix works, thanks!

PR #413, opened 2017-06-13
Fix parens for object destructuring assignment expression

### LINKED ISSUE #412: Invalid code generated for object destructuring assignment
This valid code:

```js
({ x } = { x : 1 })
```

Is printed by `recast.prettyPrint` as:

```js
{ x } = { x : 1 }
```

Which breaks the syntax. The parentheses are mandatory.

```js
> eval('({x}={x:1})')
{ x: 1 }
> x
1
> eval('{x}={x:1}')
SyntaxError: Unexpected token =
```

ef4:
Fixes #412.

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -183,13 +183,21 @@ FPp.map = function map(callback/*, name1, name2, ... */) {
 // Inspired by require("ast-types").NodePath.prototype.needsParens, but
 // more efficient because we're iterating backwards through a stack.
 FPp.needsParens = function(assumeExpressionContext) {
+  var node = this.getNode();
+
+  // This needs to come before `if (!parent) { return false }` because
+  // an object destructuring assignment requires parens for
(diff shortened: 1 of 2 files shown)


lib/fast-path.js
```diff
@@ -183,13 +183,21 @@ FPp.map = function map(callback/*, name1, name2, ... */) {
...
+  if (node.type === "AssignmentExpression" && node.left.type === 'ObjectPattern') {
+    return true;
+  }
+
   var parent = this.getParentNode();
   if (!parent) {
     return false;
   }
 
-  var name = this.getName();
```
benjamn: Might want to keep the `name` line down here, since root/parentless nodes don't have names?

ef4:
@benjamn I incorporated your feedback

benjamn:
Great, thanks so much @ef4!

Issue #440, opened 2017-10-23
parsing an input which is only a comment fails with exception when there's trailing whitespace


GerHobbelt:
See these unit tests copied below: it's the 'alt 3' one that showcases the crash (exception thrown: `"Comment location overlaps with node location"`; then when you go and hack-patch lib/comments.js to add a `return;` before that `throw` statement, you end up with a rather more sensible exception message: `"AST contains no nodes at all?"` -- which is/should-be? true for unit tests 'alt 1' and 'alt 3', but it's the trailing whitespace which makes the difference here: no trailing whitespace (alt 1) and no error report, everything hunky-dory. Add the whitespace at the end of the comment line (alt 3) and you're in big trouble... UNLESS you add a semicolon to the input to 'force' a node to exist. I'm rather stumped how to proceed... 🤔 🤔 🤔 

```javascript
    it("should correctly handle a lonesome comment (alt 1)", function () {
        var code = [
            "",
            "// boo",
            ""
        ].join(eol);

        var ast = recast.parse(code);

        assert.strictEqual(recast.print(ast).code, [
            "",
            "// boo",
            ""
        ].join(eol));
    });

    it("should correctly handle a not-so-lonesome comment (alt 2 - trailing whitespace)", function () {
        var code = [
            "",
            "// boo ",
            ";"
        ].join(eol);

        var ast = recast.parse(code);

        assert.strictEqual(recast.print(ast).code, [
            "",
            "// boo ",
            ";"
        ].join(eol));
    });

    it("should correctly handle a lonesome comment (alt 3 - trailing whitespace)", function () {
        var code = [
            "",
            "// boo ",
            ""
        ].join(eol);

        var ast = recast.parse(code);

        assert.strictEqual(recast.print(ast).code, [
            "",
            "// boo ",
            ""
        ].join(eol));
    });
```

GerHobbelt:
Note: when you also comment out that second `throw` mentioned above in lib/comments.js, all three unit tests PASS.

I don't know what the impact will be of this quick&dirty hacking, taking out two exception throwing statements, on other inputs, particularly when you go and reprint them, like I intend to do. (using recast in the jison code generator)

benjamn:
This should be fixed in `recast@0.14.3`. Thanks for reporting @GerHobbelt!

PR #441, opened 2017-10-24
Quick Patch for issue #440

GerHobbelt:
It's hacky though. See also #440: this pullreq removes those two `throw` statements. The change *seems* harmless...


benjamn:
I believe the problem you reported is fixed now (in `recast@0.14.3`), but could you rebase the tests just to make sure? I would be happy to merge them to have more regression tests.

PR #453, opened 2017-12-11
Fix bug: avoid attaching "extends" to interface declarations.

### LINKED ISSUE #452: interface declaration is broken after transform
`extends` is wrongly attached to interface declaration after transform.
The following is reproduction code.

```js
const babel = require('babel-core')
const recast = require('recast')
const assert = require('assert')

const original = `
// @flow
import type A from './a';
export interface B {};
`
const transform = () => ({
  visitor: {
    ImportDeclaration: path => path.remove()
  }
})

const { code } = babel.transform(original, {
  plugins: ["babel-plugin-syntax-flow", transform],
  parserOpts: {
    parser: recast.parse,
  },
  generatorOpts: {
    generator: recast.print,
  }
})

// Unnecessary "extends" is inserted.
assert(code === 'export interface B extends  {};')
```

I found if this line
https://github.com/benjamn/recast/blob/master/lib/printer.js#L1447
checks `n['extends'].length > 0`, this unnecessary "extends" will be disappeared.

Can you merge the PR to fix this?

shinout:
This PR fixes #452 

Although this PR would be workaround for the issue, it won't break regular cases.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1444,7 +1444,7 @@ function genericPrintNoParens(path, options, print) {
             " "
         );
 
-        if (n["extends"]) {
+        if (n["extends"] && n["extends"].length > 0) {
             parts.push(
                 "extends ",
                 fromString(", ").join(path.map(print, "extends"))


benjamn:
Thanks @shinout! Sorry for the delay.

PR #458, opened 2018-02-04
Preserve representation of numerics not in base 10

### LINKED ISSUE #457: Numeric literals that are not in base 10 always printed as base 10
I'm not 100% sure if this is by design @benjamn thoughts?

The printer uses numeric literal node's value when printing, which is the base 10 value of whatever was typed.

Eg: 

`let binary: number = 0b1010;`

Will print to:

`let binary: number = 10;`

In Esprima, the original is in the "raw" property of the literal. In Babylon, it's in extra.rawValue.

It's an easy fix and none of the existing tests rely on this, so none break when fixing it (just need to add more).

I can open a PR, but I'm not sure if this is by design or a bug.

Thanks!

Phoenixmatrix:
This fixes #457 

I still don't know if the existing behavior was by design, but if nothing else the PR can be a point of discussion. I'm also not sure if this approach is the preferred one, but I did my best to match the existing code. Esprima and Babylon handle this case pretty differently, and I had some trouble finding equivalent scenarios in the existing code.

Basically in babylon, numbers are parsed as NumericLiteral. If it's a non-base 10 number, the node has an "extra" field which contains a "raw" string, which itself contains whatever was parsed originally as is. In Esprima, the number is a plain Literal node with a raw property. I tried to make sure both cases are handled along with tests.

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -771,12 +771,23 @@ function genericPrintNoParens(path, options, print) {
     case "BigIntLiteral": // Babel 7 Literal split
         return fromString(n.value + "n");
 
+    case "NumericLiteral": // Babel 6 Literal Split
+        // keep original representation for values not in base 10
+        if (n.extra && typeof n.extra.raw === "string")
+            return fromString(n.extra.raw, options);
+
+        return fromString(n.value, options);
+
     case "BooleanLiteral": // Babel 6 Literal split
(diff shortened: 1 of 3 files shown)


Phoenixmatrix:
Also, this depends on benjamn/ast-types#247 for the ast definition change.

PR #471, opened 2018-02-20
Add better support for importKind and exportKind

bgw:
In addition to an importKind on an ImportDeclaration (`import type {Foo} from 'bar'`), it's also valid to have an importKind on an ImportSpecifier (`import {type Foo} from 'bar'`).

Additionally, it's valid to have an exportKind on an ExportNamedDeclaration (`export type {Foo};`).

This adds support for both these cases, and a couple simple tests.


bgw:
@benjamn, Thanks for maintaining recast, by the way. It makes my life much easier.

Issue #480, opened 2018-03-22
recast and babel printing two different ASTs


cangoektas:
First of all, this library is pure magic! I love it! 😍 

I discovered an interesting behaviour when using `babylon` with custom options and transforming an AST to use the experimental `export-default-from` syntax. I'm trying to transform this:

```js
export { default as foo } from 'foo';
```

into this:

```js
export foo from 'foo';
```

When I use babel to print the AST, I get what I expect. However, when using `recast.print` the result is:

```js
export { foo } from 'foo';
```

You can use the following script to reproduce the issue:

```js
const babylon = require("babylon");
const recast = require("recast");
const j = require("jscodeshift");
const generate = require("@babel/generator").default;

const recastOptions = {
  parser: {
    parse(source) {
      return babylon.parse(source, {
        sourceType: "module",
        plugins: ["exportDefaultFrom"]
      });
    }
  }
};
const code = "export { default as foo } from 'foo';";
const ast = recast.parse(code, recastOptions);

// Replace ExportSpecifier with ExportDefaultSpecifier
j(ast)
  .find(j.ExportSpecifier)
  .forEach(path => {
    const exportedName = path.node.exported.name;

    return j(path).replaceWith(
      j.exportDefaultSpecifier(j.identifier(exportedName))
    );
  });

const recastOutput = recast.print(ast).code;
const babelOutput = generate(ast).code;

console.log(recastOutput); // export { foo } from 'foo';
console.log(babelOutput); // export foo from 'foo';

```

Is there something I'm missing? Is there a way to transform the AST, so that recast prints the changes I want? Or is there something wrong with the `print` function?

Here's the `package.json` file I used:

```json
{
  "dependencies": {
    "@babel/generator": "^7.0.0-beta.42",
    "babylon": "^7.0.0-beta.42",
    "jscodeshift": "^0.5.0",
    "recast": "^0.14.5"
  }
}
```

benjamn:
Hmm, this looks like a bug in the reprinter. Specifically, `default as foo` seems to be getting replaced with just `foo`, without removing the `{...}`s. The `foo` identifier is definitely what we want, but it needs to replace more than just `default as foo` in the original code. Bugs like this can usually be solved by triggering reprinting of the entire `export` declaration.

Is there any chance you could turn this into a failing test case that I could merge into this repo? If direct AST manipulation is too much work, feel free to add `jscodeshift` as a devDependency.

cangoektas:
[...]

PR #483, opened 2018-03-23
Support the export-default-from syntax

cangoektas:
The printer has a problem with the export-default-from syntax.

Instead of printing this:

```js
export foo, { bar } from 'foo';
```

we get this:

```js
export { foo, bar } from 'foo';
```

This PR adds a failing test for that case. (#480)

@benjamn I'm not really experienced with AST's but if this doesn't require too many changes and you think a beginner could tackle it, I'd try my best if you could point me to the relevant parts of the code 😊

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -2589,6 +2589,38 @@ function printExportDeclaration(path, options, print) {
         if (decl.specifiers.length === 1 &&
             decl.specifiers[0].type === "ExportBatchSpecifier") {
             parts.push("*");
+        } else if (decl.specifiers[0].type === 'ExportDefaultSpecifier') {
+            var lastSpecifierIndex = decl.specifiers.length - 1;
+            var hasMultipleExportSpecifiers = decl.specifiers.length > 2;
+            var exportDefaultSpecifierIndex = 0;
+
+            var elements = path.map(function(specifierPath, index) {
+                if (index === exportDefaultSpecifierIndex) {
+                    return print(specifierPath);
+                }
+                if (!hasMultipleExportSpecifiers) {
+                    return concat([
+                        shouldPrintSpaces ? "{ " : "{",
+                        print(specifierPath),
+                        shouldPrintSpaces ? " }" : "}"
+                    ]);
+                }
+                if (index === 1) {
+                    return concat([
+                        shouldPrintSpaces ? "{ " : "{",
+                        print(specifierPath)
+                    ]);
+                }
+                if (index === lastSpecifierIndex) {
+                    return concat([
+                        print(specifierPath),
+                        shouldPrintSpaces ? " }" : "}"
+                    ]);
+                }
+                return concat([print(specifierPath), ", "]);
+            }, "specifiers");
+
+            parts.push(fromString(", ").join(elements));
         } else {
             parts.push(
                 shouldPrintSpaces ? "{ " : "{",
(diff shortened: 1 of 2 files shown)


benjamn:
If you want to give it a try, I would run `npm run debug` to enable the inspector, and then set a breakpoint in `printExportDeclaration` [here](https://github.com/benjamn/recast/blob/1648a2147e1d64f44635172e417d547094d3bf58/lib/printer.js#L2568) to see what happens.

Now that I think about it, there's a good chance `printExportDeclaration` isn't even getting called in this case, because the specifier child nodes are getting reprinted in place, so you might want to set a breakpoint in the `ExportDefaultSpecifier` printer case [here](https://github.com/benjamn/recast/blob/1648a2147e1d64f44635172e417d547094d3bf58/lib/printer.js#L506) as well.

If the child specifier is getting reprinted by itself, without triggering reprinting of the parent `export` declaration, then my guess would be that `canReprint` is `true` on [this line](https://github.com/benjamn/recast/blob/1648a2147e1d64f44635172e417d547094d3bf58/lib/patcher.js#L323) in `findArrayReprints`, since `exportDecl.specifiers` is an array, and the `ExportDefaultSpecifier` is an element of that array.

So here's my hint: if you can get `canReprint` to be `false` when a specifier changes types (from `ExportSpecifier` to `ExportDefaultSpecifier`), then `findArrayReprints` will return `false` and the whole `export` declaration should get reprinted, resulting in the desired output.

cangoektas:
Hey @benjamn, thanks a lot for the hints!

It turns out the `printExportDeclaration` function didn't handle the case with the `ExportDefaultSpecifier` and would always put curly braces around the specifiers.

I was able to fix this but I think it's a bit of a stupid implementation. I didn't fully get the concept of Lines and how to handle the FastPath objects. If you have any ideas on how to improve this, I'm all ears :)

benjamn:
@cangoektas I think this is a great start, and a correct diagnosis of the problem. If you want to check your implementation of `printExportDeclaration`, have a look at [the printer case for `ImportDeclaration`](https://github.com/cangoektas/recast/blob/cd35847148ea17c191747f4b72b84823be7d6e19/lib/printer.js#L511), which deals with some of the same issues (braced and unbraced specifiers).

cangoektas:
Hey @benjamn, sorry I couldn't get back earlier. I plan on continuing with this on the weekend. One thing I didn't understand fully with the `ImportDecleration` is this part: https://github.com/cangoektas/recast/blob/cd35847148ea17c191747f4b72b84823be7d6e19/lib/printer.js#L543

Does it handle the case where imports span across multiple lines? And recast tries to preserve that formatting? Is there a test for such cases that I can look up?

benjamn:
Right, that code checks if the line would be "too long" if printed without line breaks, and adds line breaks if so. Once this logic is executing (because this part of the AST node was sufficiently modified), the `ImportDeclaration` is going to be reprinted, according to the rules of the pretty-printer, so there's no special effort to infer or preserve the original style. This code is just trying to do something reasonable depending on how many specifiers there are.

cangoektas:
Hey @benjamn, I think I got it. Also added another export statement to cover that case with line-breaks. Does this look good to you?

cangoektas:
Hey @benjamn, did you had time to review this? Is there anything else I can do?

PR #491, opened 2018-04-23
Fix #490

### LINKED ISSUE #490: printing empty export declaration
recast's pretty-printer doesn't print `export {}` nodes correct.  (that is, export declarations with zero export specifiers)  It prints `export ` and omits the empty curly braces.

# Motivation

I want to use recast to automatically add `export {}` to a bunch of files.  This causes my IDE to consider them all external modules (not scripts) and flag certain errors it otherwise would not.

# The codemod

```
const emptyExport = builders.exportNamedDeclaration(null, [], null);
ast.program.body.unshift(emptyExport);
```

Recast's pretty-printer prints the node as `export ` instead of `export {}`

Relevant logic is here: https://github.com/benjamn/recast/blob/master/lib/printer.js#L2586-L2587

It skips over all nodes with a zero-length specifiers array.

My workaround is `builders.identifier('export {}');`

cspotcode:


lib/printer.js
```diff
@@ -2583,8 +2583,7 @@ function printExportDeclaration(path, options, print) {
     if (decl.declaration) {
         parts.push(path.call(print, "declaration"));
 
-    } else if (decl.specifiers &&
-               decl.specifiers.length > 0) {
+    } else if (decl.specifiers) {
```
benjamn: This change makes sense, but can we also make `shouldPrintSpaces` false if `decl.specifiers.length === 0`, to avoid the extra spaces between the braces?

benjamn:
Those test failures are not your fault. I will merge this PR after I've fixed those tests on `master`.

cspotcode:
@benjamn  I made an attempt in #503 and then realized one of the failures is a real recast bug, reprinting arrow functions where the body is an expression that starts with a curly brace.  I opened #504 to track it.

PR #494, opened 2018-04-24
Add a printer case for flow tuple type annotations

### LINKED ISSUE #495: Unprintable type: "TupleTypeAnnotation"
I'm using recast to code gen some flow library definitions, and couldn't print a `TupleTypeAnnotation`. I saw #422 which was closed by the issue-opener, but would love to be able to print these!

I'm not super familiar with the code, but I've added a printer case and test case for it in #494, if it looks good to you @benjamn!

ben-chin:
Closes #495

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -1391,7 +1391,6 @@ function genericPrintNoParens(path, options, print) {
     case "NamedSpecifier":
     case "Comment": // Supertype of Block and Line.
     case "MemberTypeAnnotation": // Flow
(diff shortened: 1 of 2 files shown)


benjamn:
@ben-chin Thanks for this, it looks great! As I mentioned in https://github.com/benjamn/recast/pull/491#issuecomment-384109767, I need to fix the failing tests on `master` first, but I will merge this PR as soon as I do that.

ben-chin:
thanks for fixing the tests @benjamn!

PR #501, opened 2018-05-18
[WIP] Fix 493: allow controlling line terminator of parsed AST (affects node offsets)

### LINKED ISSUE #493: Unable to control *parsed* lineTerminator
recast doesn't let me control the lineTerminator used for the *parsed* AST.  This is significant because it subtly affects location information within the AST.

I'm using TypeScript's language service to facilitate some advanced codemods.  I get a Babylon node from Recast and use it's location information to retrieve the corresponding node in TypeScript's AST.  This gives me access to semantic information, a list of references to a variable or property across the entire codebase, advanced diagnostics, etc.

However, since I'm running on Windows, Recast is converting all LF line-endings to CRLF before parsing via Babylon.  This introduces a +1 offset to all Babylon node locations.

(TypeScript's API use character offset locations, not line + column #)

My workaround is to normalize all code to CRLF before passing it to the TS language service.  But this isn't ideal, because although locations in both ASTs match, neither matches the files on disk.

I think this code is the culprit.  It doesn't pass a `lineTerminator` option, so the generated string contains CRLF terminators on Windows.
https://github.com/benjamn/recast/blob/master/lib/parser.js#L21-L25

cspotcode:
Can I get a quick code review on this?  I had to hook into the esprima parser for the unit tests, so I'd like to potentially refactor it into whatever style you prefer.

Fixes #493.

### DIFF
--- a/lib/parser.js
+++ b/lib/parser.js
@@ -20,6 +20,7 @@ exports.parse = function parse(source, options) {
 
   const sourceWithoutTabs = lines.toString({
     tabWidth: options.tabWidth,
+    lineTerminator: options.lineTerminator,
     reuseWhitespace: false,
     useTabs: false
   });
--- a/parsers/esprima.js
+++ b/parsers/esprima.js
@@ -10,6 +10,11 @@
 const getOption = require("../lib/util.js").getOption;
 
 exports.parse = function (source, options) {
+  return exports.__parseInternal(source, options);
+};
+
+// required by unit tests to pass additional options to esprima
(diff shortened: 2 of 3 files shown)


cspotcode:
@benjamn I rebased this to the latest master and moved it to a new `parsedLineTerminator` option, distinct from `lineTerminator`.  So this should be fully backwards compatible; it'll only take effect if a consumer sets the new option.

The failing test seems unrelated, since it's failing across the board on all new greenkeeper PRs: https://travis-ci.org/benjamn/recast/jobs/494081987

PR #502, opened 2018-05-18
Fix unit tests that use Function.prototype.toString()

cspotcode:
Function.prototype.toString() returns the line endings of the source code on disk, which is not guaranteed to match `os.EOL` without a `.gitattributes` file.

The fix is simple: normalize line endings after calling `toString()` to get predictable results.

### DIFF
--- a/test/lines.js
+++ b/test/lines.js
@@ -14,6 +14,12 @@ function check(a, b) {
     }));
 }
 
+// Function.prototype.toString returns the same newlines as the source on disk
+// (possibly \r\n, possibly \n even on Windows), so normalize them.
+function functionToString(fn) {
+    return fn.toString().replace(/\r?\n/g, eol);
+}
+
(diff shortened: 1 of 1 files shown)


test/lines.js
```diff
@@ -272,7 +276,7 @@ describe("lines", function() {
     });
 
     it("Slice", function SliceTest() {
-        var code = String(SliceTest),
+        var code = functionToString(SliceTest);
```
benjamn: We should keep the comma here.

test/lines.js
```diff
@@ -294,7 +298,7 @@ describe("lines", function() {
     }
 
     it("GetSourceLocation", function GetSourceLocationTest() {
-        var code = String(GetSourceLocationTest),
+        var code = functionToString(GetSourceLocationTest);
```
benjamn: And here.

PR #511, opened 2018-05-30
Implement printer cases for Optional{Member,Call}Expression.

benjamn:
https://github.com/benjamn/ast-types/pull/272
https://github.com/tc39/proposal-optional-chaining

cc @qfox @cpojer

### DIFF
--- a/lib/printer.js
+++ b/lib/printer.js
@@ -267,13 +267,16 @@ function genericPrintNoParens(path, options, print) {
         ]);
 
     case "MemberExpression":
+    case "OptionalMemberExpression":
         parts.push(path.call(print, "object"));
 
         var property = path.call(print, "property");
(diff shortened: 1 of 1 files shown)


qfox:
I would say this is missing some test cases but the code is looking good.

Basically the optional calls are `foo?.(foo)`  according to the spec (https://github.com/tc39/proposal-optional-chaining). Super ugly, but yeah. So this looks good to me.

benjamn:
Since this proposal is only at Stage 1, I predict there will be breaking changes before the syntax is finalized, but that's just how it goes with ECMAScript standards. Recast and ast-types will adapt as necessary, and may even be useful for migrating from old syntax to new syntax.

benjamn:
Ok, this has been published as part of `recast@0.15.0`.

Issue #516, opened 2018-07-04
Printer outputs invalid code for default export of IIFE


robpalme:
Constructing an `export default` of an anonymous IIFE causes the printer to output the following invalid JS.

```javascript
export default function() {}();
```

Desired output is:

```javascript
export default (function() {}());
```

benjamn:
Interesting! The way we usually solve parenthesization issues is by adding rules to `FastPath#needsParens`, if you have time to take a look: https://github.com/benjamn/recast/blob/7f7e76805ba898c5f637e5a27e39830a2c7e221e/lib/fast-path.js#L228

PR #517, opened 2018-07-04
Fix printing of IIFEs as default exports

robpalme:
Previously the code forgot to add sufficient parens.

fixes issue #516

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -407,6 +407,13 @@ FPp.needsParens = function(assumeExpressionContext) {
       return true;
     }
 
+  case "CallExpression":
+    if (n.ExportDefaultDeclaration.check(parent) &&
+        name === "declaration" &&
+        n.FunctionExpression.check(node.callee)) {
+      return true;
+    }
+
   default:
     if (parent.type === "NewExpression" &&
         name === "callee" &&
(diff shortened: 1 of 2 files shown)


lib/fast-path.js
```diff
@@ -407,6 +407,13 @@ FPp.needsParens = function(assumeExpressionContext) {
       return true;
     }
 
+  case "CallExpression":
+    if (parent.type === "ExportDefaultDeclaration" &&
+        name === "declaration" &&
+        node.callee.type === "FunctionExpression") {
```
benjamn: The test failures seem to be suggesting that `node.callee` could be undefined here, though I'm honestly not sure how that could be the case since `node.type === "CallExpression"`…

robpalme: For some reason `export default {}` in `es6test.js` hits this `CallExpression` case yet clearly contains no calls.  That seems like a pre-existing orthogonal issue.

Using more robust node checking fixes this test case.

lib/fast-path.js
```diff
@@ -407,6 +407,13 @@ FPp.needsParens = function(assumeExpressionContext) {
       return true;
     }
```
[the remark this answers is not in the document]
benjamn: Fixed in 083076d974b14b1427d3469d5f5cd8f6f6943623

benjamn:
This change has been published in `recast@0.15.2`, thanks @robpalme!

PR #535, opened 2018-08-29
Support tsx files with exported typescript parser

evans:
Adding the jsx plugin provides the ability to parse typescript files containing jsx syntax. See https://github.com/facebook/jscodeshift/issues/180#issuecomment-399618290 for more color.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,31 +5,31 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.54",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.54.tgz",
-      "integrity": "sha1-ACT5b99wKKIdaOJzr9TpUyFKHq0=",
+      "version": "7.0.0",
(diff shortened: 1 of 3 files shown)


benjamn:
Thanks for this, @evans! Those test failures seem relevant, though. I had thought that JSX conflicted with some TypeScript `<type>` syntax, or I would have enabled it by default.

eventualbuddha:
This seems unlikely to move forward, so I’m closing it. Consumers of recast can easily add the jsx plugin if they like.

kube:
It's a little bit sad to see that this was closed.

TSX has some syntax conflicts with generic type parameters (e.g. `const fn = <T>(t: T) => t`), but it is worked around in TypeScript by specifying extension (`.tsx` vs `.ts`).

Maybe we could reopen the PR, and just define it as a new parser.

I know that I can define plugins manually, but it was really close here.

For the parser name IMO it should be `typescriptReact`:

Screenshot 2022-04-19 at 03 02 02

kube:
Sorry, after trying, indeed there are collisions between syntaxes as the current PR defines plugins.

PR #536, opened 2018-09-10
Pass `range` option along to esprima parser.

fkling:
https://github.com/fkling/astexplorer/issues/334 mentions how syntax highlighting doesn't work for recast by default. So I noticed that the `range` option isn't passed along to esprima. 

We could of course also make it `true` by default.


benjamn:
Since Recast uses only `.loc` and does not need the range information for its own purposes, I think defaulting to false is fine for now, but I'm not opposed to flipping the default if that satisfies a use case.

Issue #533, opened 2018-08-12
Transforming 


Retsam:
I believe this to be largely the same underlying causes as #327 , so I suspect there's nothing to be done about it, at the moment, but I thought it'd be worth opening an issue to track it nonetheless.

benjamn:
I think you're right about the similarity to #327, but I'm happy to keep this issue open because you've found a slightly different test case.

gnprice:
It looks like this issue no longer reproduces.

Here's the output at the given AST Explorer link, at recast 0.21.1:
`({ foo } = bar)`

So that's the one necessary layer of parentheses (if it were just `{ foo } = bar`, the leading `{` wouldn't introduce an expression at all but instead a BlockStatement, and then the `=` would be a syntax error), but no unnecessary additional layer.

Issue #528, opened 2018-08-03
Bug: parenthesis are not preserved in unary arrow function


yuriyyakym:
Let's say we have an arrow function like this
`const identity = (a) => a;`
after parsing it and then printing ast back, parentheses will be lost:
`const identity = a => a;`.

The logics is described here: https://github.com/benjamn/recast/blob/master/lib/printer.js#L369
We need to check whether parentheses were present in the input and add them in the output if were present.

It is fixed in https://github.com/benjamn/recast/pull/527

gnprice:
This is expected behavior.

If you parse that code and then print it back with `printer.print`, it does in fact come back verbatim. For example:
```
$ node
> recast = require('recast'); source = 'const identity = (a) => a;';
> recast.print(recast.parse(source)).code === source
true
```

One can also demonstrate this by adding `check("const identity = (a) => a;");` inside any of the tests in `test/parens.ts`.

---

The test case added in #529 printed the AST instead with `printer.printGenerically`. The job of `printGenerically` is that it doesn't care about the details of how things were formatted in input. That function itself isn't actually part of the API, but it's exposed in the API as `recast.prettyPrint`, which has the following documentation:
```js
/**
 * Print without attempting to reuse any original source code.
 */
export function prettyPrint(node: types.ASTNode, options?: Options) {
  return new Printer(options).printGenerically(node);
```

So:
```
> recast.prettyPrint(recast.parse(source)).code
'const identity = a => a;'
```
which is fine and expected.

Issue #513, opened 2018-06-28
Misplaced parenthesis in object destructuring argument with trailing argument comma


tlrobinson:
I haven't been able to pin down exactly why this is happening, but here's my smallest reproduction so far. I'm using `flow` the parser.

```javascript
class Foo {
  method = item => {
    return (
      <div className=""></div>
    );
  };
}

function x({ className, }) {
}
```

gets printed as

```javascript
class Foo {
  method = item => {
    return (
      <div className=""></div>
    );
  };
}

function x({ (className, })) {
}
```

Note the extra (and mis-nested) parenthesis in the last function.

Interestingly, changing just about anything about the code above causes the problem to go away.

benjamn:
This is just a guess, but in the past, bugs like this have sometimes been due to the parser providing slightly incorrect `.loc.{start,end}.{line,column}` information. Like, perhaps the parser gets confused by the trailing comma and decides that the `className` identifier location extends all the way past the end of the parameter list… or something like that.

benjamn:
Can you still reproduce this? I've been trying to write a regression test, and it seems to work now. If my previous theory is correct, the Flow parser might have fixed their `.loc` tracking in the meantime.

Issue #512, opened 2018-06-25
Unexpectedly changing the code when encountering parentheses in arrow function body


kevin940726:
Related to #504, #505.

Recast version: `0.15.0`

As @cspotcode fixed a issue in #504, now recast will unexpectedly transform the code while it shouldn't.

Consider the following code

```js
const x = () => ({})["foo"];
```

Running with recast will now print the following code back

```js
const x = () => ({}["foo"]);
```

Note that the ending parentheses position has changed and therefore changing the source code unexpectedly. This will not happen when running with `recast@0.14.7`.

Any idea how to fix it?

kevin940726:
I think it's already been fixed

PR #537, opened 2018-09-10
Use node.loc.tokens to improve handling of parentheses.

benjamn:
Recast has suffered for a long time because it did not have reliable access to the lexical analysis of source tokens during reprinting, and instead relied on scanning source characters, which is error-prone because identifying comments, string literals, regular expression literals, and other large tokens is difficult/impossible without doing a full tokenization and sometimes a small amount of initial parsing.
    
Most importantly, accurate token information can be used to detect whether a node was originally wrapped with parentheses, even if those parentheses were separated from the node by comments or other incidental non-whitespace text, such as trailing commas. Here are just some of the issues that have resulted from the lack of reliable token information:

- [x] #533
- [ ] #528
- [ ] #513
- [x] #512
- [x] #504
- [x] #366
- [x] #327
- [ ] #286
    
With this change, every node in the AST returned by `recast.parse` will now have a `node.loc.tokens` reference to shared array containing the entire sequence of original source tokens, as well as `node.loc.{start,end}.token` indexes into this array of tokens, such that
```js
node.loc.tokens.slice(
  node.loc.start.token,
  node.loc.end.token
)
```
returns a complete list of all source tokens contained by the node. Note that some nodes (such as comments) may contain no source tokens, in which case `node.loc.start.token === node.loc.end.token`, which will be the index of the first token *after* the position where the node appeared.
    
Most parsers can expose token information for free / very cheaply, as a byproduct of the parsing process. In case a custom parser is provided that does not expose token information, we fall back to Esprima's tokenizer. While there is considerable variation between different parsers in terms of AST format, there is much less variation in tokenization, so the Esprima tokenizer should be adequate in most cases (even for JS dialects like TypeScript). If it is not adequate, the caller should simply ensure that the custom parser exposes an `ast.tokens` array containing token objects with `token.loc.{start,end}.{line,column}` information.

### DIFF
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -216,7 +216,13 @@ exports.getReprinter = function(path) {
 
     // Recall that origLoc is the .loc of an ancestor node that is
     // guaranteed to contain all the reprinted nodes and comments.
-    return patcher.get(origLoc).indentTail(-orig.loc.indent);
+    const patchedLines = patcher.get(origLoc).indentTail(-orig.loc.indent);
+
+    if (path.needsParens()) {
+      return linesModule.concat(["(", patchedLines, ")"]);
+    }
+
+    return patchedLines;
   };
 };
 
--- a/lib/patcher.js
+++ b/lib/patcher.js
@@ -190,10 +190,9 @@ exports.getReprinter = function(path) {
         patcher.deleteComments(oldNode);
       }
 
-      var newLines = print(
-        reprint.newPath,
-        needToPrintNewPathWithComments
-      ).indentTail(oldNode.loc.indent);
+      var newLines = print(reprint.newPath, {
+        includeComments: needToPrintNewPathWithComments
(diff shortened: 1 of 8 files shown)


lib/patcher.js
```diff
@@ -487,29 +427,11 @@ function findChildReprints(newPath, oldPath, reprints) {
...
-  // CallExpression passes the hasParens test. If we relied on the
-  // path.needsParens() && !hasParens(oldNode) check below, the absence of
-  // a closing parenthesis after the FunctionExpression would trigger
-  // pretty-printing unnecessarily.
-  if (Node.check(newNode) &&
-      !newPath.canBeFirstInStatement() &&
-      newPath.firstInStatement() &&
-      !hasOpeningParen(oldPath)) {
-    return false;
-  }
```
benjamn: Very glad to see this ceremony moved entirely into `FastPath#hasParens`.

PR #554, opened 2018-11-20
Convert repo to TypeScript

brieb:
Continuation of https://github.com/benjamn/ast-types/pull/300

Reviewer: @benjamn


lib/lines.ts
```diff
@@ -656,6 +699,10 @@ Lp.slice = function(start, end) {
     end = this.lastPos();
   }
 
+  if (!start) {
+    throw new Error("cannot slice with end but not start");
+  }
```
brieb: Needed this type guard in order for TS to know `start` is present later on in this function

brieb:
Hm. Looks like babel no longer supports node 4 and 5, which is why the build is failing.

> Support for Node.js 0.10, 0.12, 4 and 5 has been dropped
https://babeljs.io/docs/en/v7-migration

Are we okay with also removing support for 4 and 5?

benjamn:
Would we be removing support for Node 4 and 5 when *using* `recast`, or just when building the package using Babel in development?

brieb:
In theory, we would just need it for building and running the tests, though I'm not sure if the transpilation process would emit code only runnable in the versions of node Babel officially supports.

brieb:
Thought about this more and it feels pretty silly to introduce all this babel overhead in the build pipeline just in order to use babel-plugin-add-module-exports. I took a look at the [implementation](https://github.com/59naga/babel-plugin-add-module-exports) and it's pretty straightforward. So, I wrote a [script](https://github.com/brieb/ts-add-module-exports) that post-processes the tsc emitted files to do the same transformation (using recast internally, of course <3). This lets us use tsc for both the declarations and js transpilation (which also conveniently helps us get around [limitations in babel when re-exporting types](https://github.com/babel/babel/issues/8361)). That greatly simplified the build setup in this PR and will hopefully make this build failure go away. I'll push up a commit that changes the build setup over to that so you can check it out.

brieb:
Added printer support for the types [I introduced](https://github.com/benjamn/recast/blob/6a7fa9b8e5f720051e636962ab5e1d84690d591a/lib/printer.ts#L2363-L2369).

Used https://github.com/babel/babel/tree/master/packages/babel-generator/src/generators as a reference for the implementation

benjamn:
Do you think you might have time to rebase this PR to remove the `[TEMPORARY]` commits that updated prerelease versions of `ast-types`, and possibly the Babel 7 compilation stuff, since you ended up switching to pure TypeScript? If you don't have time, I can handle it.

brieb:
Will do!

lib/printer.ts
```diff
@@ -844,6 +871,9 @@ function genericPrintNoParens(path, options, print) {
     case "DirectiveLiteral": // Babel 6
         return fromString(nodeStr(n.value, options));
 
+    case "InterpreterDirective":
+        return fromString(`#!${n.value}\n`);
```
benjamn:
```suggestion
        return fromString(`#!${n.value}\n`, options);
```
The potential values of `n.value` seem diverse enough to potentially be affected by the options.

lib/comments.ts
```diff
@@ -1,19 +1,16 @@
...
-var childNodesCacheKey = require("private").makeUniqueKey();
+import { Lines, concat } from "./lines";
+import { comparePos, fixFaultyLocations } from "./util";
+import { makeUniqueKey } from "private";
+var childNodesCacheKey = makeUniqueKey();
 
 // TODO Move a non-caching implementation of this function into ast-types,
 // and implement a caching wrapper function here.
-function getSortedChildNodes(node, lines, resultArray) {
+function getSortedChildNodes(node: any, lines: any, resultArray?: any) {
```
benjamn: Is there some automated transform that helps with this, or was this all by hand? 😬

brieb: Did these ones by hand, but I'm working on a tool that automates patterns like this. It's called ts-migrate (super alpha at this point) and its a collection of codemods and scripts for migrating codebases to TS. Here's the code that would have added these https://github.com/brieb/ts-migrate/blob/dc324e1aefa6952c78b3bd092ca4b34ff8eb64c4/src/plugins/explicit-any/withExplicitAny.ts

test/typescript.ts
```diff
@@ -1,24 +1,17 @@
...
+import fs from "fs";
+import recast from "../main";
+import types from "../lib/types";
+import {EOL as eol} from "os";
+import * as parser from "../parsers/typescript";
+
+// Babel 7 no longer supports Node 4 or 5.
+var nodeMajorVersion = parseInt(process.versions.node, 10);
+(nodeMajorVersion >= 6 ? describe : xdescribe)
+("TypeScript", function() {
```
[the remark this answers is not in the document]
brieb: Copied that pattern from here https://github.com/benjamn/recast/blob/f43bbc06c0150c27483de78503f5cb93e07a5727/test/jsx.js#L8-L9

lib/comments.ts
benjamn: Cool!

PR #556, opened 2018-11-28
Fix issue with export default template literals

### LINKED ISSUE #414: Bug - Can't Export Default Template String
First off, this library is awesome, so thank you! 😄 Using it with `jscodeshift`.

#### Description
[`export default`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) allows any expression to be exported, but exporting a template string causes a TypeError in `recast`.
#### Code Snippet
```
const ast2 = recast.parse("export default `hello world`;");
// Error: Uncaught TypeError: Cannot read property 'start' of null at fixTemplateLiteral
```
#### Reproduce
https://esnextb.in/?gist=f49ee2da9306734bb4813ebcff734c38
#### Fix
Looks like the line of code where the TypeError gets thrown is accessing `node.loc.start` in [`fixTemplateLiteral`](https://github.com/benjamn/recast/blob/98599d1a36a686919b5fcf47616884bad9ac790a/lib/util.js#L262).

ganemone:
This fixes an issue where exporting a template literal as default caused the parser to fail. It also adds a regression test.

Fixes #414

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -265,25 +265,28 @@ function fixTemplateLiteral(node, lines) {
     // If there are no quasi elements, then there is nothing to fix.
     return;
   }
+  
+  // node.loc is not present when using export default with a template literal
+  if (node.loc) {
+    // First we need to exclude the opening ` from the .loc of the first
+    // quasi element, in case the parser accidentally decided to include it.
+    var afterLeftBackTickPos = copyPos(node.loc.start);
+    assert.strictEqual(lines.charAt(afterLeftBackTickPos), "`");
+    assert.ok(lines.nextPos(afterLeftBackTickPos));
+    var firstQuasi = node.quasis[0];
+    if (comparePos(firstQuasi.loc.start, afterLeftBackTickPos) < 0) {
+      firstQuasi.loc.start = afterLeftBackTickPos;
+    }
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -265,25 +265,28 @@ function fixTemplateLiteral(node, lines) {
     // If there are no quasi elements, then there is nothing to fix.
     return;
   }
+  
+  // node.loc is not present when using export default with a template literal
```
benjamn: Any idea why `node.loc` would be missing?

ganemone: Not sure to be honest. I can reproduce this with the regression test however. I also can reproduce this with `@babel/parser`, even with `ranges` turned on.

ganemone:
Any reservations on merging this?

lib/util.js
benjamn: Ahh, that's something we do in a few places to disable reprinting problematic nodes, e.g. [here](https://github.com/benjamn/recast/blob/45773b02ad1dfbd61ac0050df17587d0d83dca7e/lib/util.js#L186-L188).

This PR is definitely safe/fine for now. Sorry for the wait!

PR #562, opened 2019-01-12
Convert Lines to class.

benjamn:
When everything was converted to TypeScript, we intentionally held off on converting some existing constructor functions to proper classes.

This PR converts the `Mapping` and `Lines` types to classes.

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -5,3 +5,16 @@
 import astTypes from "ast-types";
 export default astTypes;
 export * from "ast-types";
+
+// TODO Get these types from ast-types.
+
+export type Position = {
+  line: number;
+  column: number;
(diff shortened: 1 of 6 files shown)


lib/types.ts
```diff
@@ -5,3 +5,16 @@
 import astTypes from "ast-types";
 export default astTypes;
 export * from "ast-types";
+
+// TODO Get these types from ast-types.
+
+export type Position = {
```
benjamn: In order to obtain these types from `ast-types`, we need to be a little more careful about making optional fields optional, rather than just giving them a default value (e.g. the `source` field of `SourceLocation`, which a non-optional `string | null` according to `ast-types`).

PR #551, opened 2018-10-25
Inexact printing

jbrown215:
Flow added an inexact flag to the ObjectTypeAnnotation node. When it is true, the object should have a `...` at the end of its props list. See  https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf

PR #577, opened 2019-02-21
Fix export all declarations to include a trailing semicolon.

eventualbuddha:


eventualbuddha:
@benjamn any idea why these failures are occurring? They seem unrelated to the changes.

benjamn:
@eventualbuddha Yep, totally unrelated, and fixed in #551 (I hope 🤞). The `@babel/parser` TypeScript implementation apparently has trouble with [this line](https://github.com/babel/babel/blob/48d66eb6482a6bc029a884f113e78cbfe8183e2e/packages/babel-parser/test/fixtures/typescript/arrow-function/optional-parameter/input.js#L1), since it's just a function expression and not a declaration/statement.

PR #588, opened 2019-04-14
strictMode option for @babel/parser

sebdeckers:
Adding the strictMode option. This option is used by `recast/parsers/babel`. It was not being passed along by `recast.parse(source, { strictMode })`.


sebdeckers:
FWIW the strictMode option is probably not needed anymore, or at least should default to `true`. See: https://github.com/babel/babel/pull/9725#pullrequestreview-217028751

eventualbuddha:
Without a clear idea of what this is fixing, I’m inclined to leave it as-is. Let us know if you have an example where this helps.

PR #591, opened 2019-04-28
Update ast-types to v0.12.4

devongovett:


devongovett:
Oops. Looks like this is already covered by #590.

PR #593, opened 2019-05-13
Modernize and TypeScript-ify the main.ts entry point module.

benjamn:
Using `Object.defineProperties` to populate the `exports` object in `main.ts` is an ancient holdover from CommonJS. Now that this repository uses TypeScript (#554), we should embrace TS/ESM syntax as fully as possible.

From an external perspective, the most important change here is that the `recast` package no longer has a `default` export, and now uses only named exports.

Some exported types (`ASTNode`, `NodePath`, and `Type`) are no longer exported directly from `main.ts`, but may instead be obtained via `recast.types.{ASTNode,NodePath,Type}`.

These changes definitely necessitate a minor version bump, at the very least.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1710,26 +1710,6 @@
         "once": "^1.3.1"
       }
     },
-    "recast": {
-      "version": "0.16.2",
-      "resolved": "https://registry.npmjs.org/recast/-/recast-0.16.2.tgz",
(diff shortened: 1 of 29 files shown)


package-lock.json
```diff
@@ -1710,26 +1710,6 @@
         "once": "^1.3.1"
       }
     },
-    "recast": {
-      "version": "0.16.2",
```
benjamn: Nice to no longer be depending on ourselves!

PR #609, opened 2019-06-27
fix: AssignmentPattern in destructuring

### LINKED ISSUE #608: Missing AssignmentPattern in a newly created node
When I create new node using `babel` as a `pareser`, make some transformation, adding `AssingmentPattern` I can't see it in a `print` result.

```js
const { parse, print } = require("recast");
const template = require('@babel/template').default;

const source = 'var {x} = y';

const ast = parse(source, {
    parser: require("recast/parsers/babel")
});

const newNode = template.ast(`
  const {myModule = 'abc'} = require("my-module");
`);

traverse(ast, {
    VariableDeclaration(path) {
        path.replaceWith(newNode);
    }
});

console.log(print(ast).code);
console.log('babel: ', generate(ast).code);
```
This code outputs:

```js
const {
  myModule
} = require("my-module");

babel:  const {
  myModule = 'abc'
} = require("my-module");
```

`AssignmentPattern` is missing. Looks like the problem is in [printer](https://github.com/benjamn/recast/blob/master/lib/printer.ts#L764-L766), when there is no shorthand, `value` property do not processed at all.

As you see `babel generate` works good, so it's not a `babel` bug.

coderaiser:
Fixes #608.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -755,8 +755,12 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         }
 
         var key = path.call(print, "key");
+        var value = path.call(print, "value");
+
         if (n.computed) {
             parts.push("[", key, "]");
+        } else if (n.shorthand && n.value.type === 'AssignmentPattern') {
+            parts.push(value);
         } else {
             parts.push(key);
         }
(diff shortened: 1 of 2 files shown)


coderaiser:
Looks like this already fixed in master https://github.com/benjamn/recast/commit/8f4e5c83167a73327f0113c31698af31d8791b17.

PR #613, opened 2019-07-09
Fix printing expressions with type parameters

zxbodya:
added code to print `typeParamentes` for:
 - FunctionExpression (case when it is anonymous function)
 - NewExpression
 - CallExpression

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -928,6 +928,9 @@ function genericPrintNoParens(path: any, options: any, print: any) {
 
     case "NewExpression":
         parts.push("new ", path.call(print, "callee"));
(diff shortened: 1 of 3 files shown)


goodmind:
@zxbodya you forgot `typeArguments` for Flow

zxbodya:
@goodmind  Sorry for misleading PR title - actually it was only about typescript (caught it while testing tool converting flow syntax to typescript). 

But yes, it makes sense to make a fix for flow as well - added it to PR.

steelbrain:
Thanks for merging this @benjamn! Could we please get a release out (soon) with this included?

PR #617, opened 2019-08-01
Add semicolons after directives only as needed.

eventualbuddha:
This description will use the following code for illustration:

```js
(function () {
  "use strict";
  hello;
})();
```

In the situation where the containing `BlockStatement` changed but the directive itself did not, recast would simply print the directive as-is while generically printing the block. Since the `Directive` node `loc` includes the trailing semicolon, that part gets printed as `"use strict";`. Subsequently, the `BlockStatement` printer would unconditionally add a semicolon part after each directive string. This caused recast to print a double semicolon in such circumstances, i.e. `"use strict";;`.

Originally reported in https://github.com/codemod-js/codemod/issues/157.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -608,10 +608,8 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         if (n.directives) {
             path.each(function(childPath: any) {
                 parts.push(
-                    print(childPath).indent(options.tabWidth),
-                    ";",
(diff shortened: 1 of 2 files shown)


eventualbuddha:
@benjamn I decided to just merge it since it seemed unlikely to get anyone else to review. Mind doing a (likely patch) release?

PR #624, opened 2019-09-01
parenthesis around arrow function body when returning object literal

zxbodya:
Add parenthesis around arrow function body when returning object literal with typescript typecast

```ts
() => ({} as object);
```


eventualbuddha:
Fixed in v0.18.4.

PR #642, opened 2019-10-16
refactor: use block scoping

eventualbuddha:
This was done automatically using the following:

```
npm i @codemod/cli @resugar/codemod-declarations-block-scope
./node_modules/.bin/codemod -p @resugar/codemod-declarations-block-scope {src,test}/**/*.ts
```

@benjamn, if there's some reason not to do this let me know. Otherwise I'll merge in a few days.

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -164,7 +164,7 @@ interface DeprecatedOptions {
   esprima?: any;
 }
 
-var defaults: Options = {
+const defaults: Options = {
   parser: require("../parsers/esprima"),
   tabWidth: 4,
   useTabs: false,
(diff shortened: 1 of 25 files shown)


eventualbuddha:
Also, a fun note about this PR: it's recast rewriting recast.

lib/options.ts
```diff
@@ -164,33 +164,34 @@ interface DeprecatedOptions {
...
+        tolerant: true,
+        quote: null,
+        trailingComma: false,
+        arrayBracketSpacing: false,
+        objectCurlySpacing: true,
+        arrowParensAlways: false,
+        flowObjectCommas: true,
+        tokens: true
+      },
+      hasOwn = defaults.hasOwnProperty;
```
benjamn: I think I see what happened here (lining up the variables vertically, then indenting the object literal by two spaces relative to that), but I wonder if this is really the indentation we want?

eventualbuddha: Probably not. I'll split into two declarations and that should fix it.

PR #643, opened 2019-10-16
style: reformat `lib/` using prettier

eventualbuddha:
I omitted `test/` because there's quite a few vertical alignments of sample code that prettier does not preserve.

### DIFF
--- a/lib/comments.ts
+++ b/lib/comments.ts
@@ -22,14 +22,12 @@ function getSortedChildNodes(node: any, lines: any, resultArray?: any) {
   fixFaultyLocations(node, lines);
 
   if (resultArray) {
-    if (n.Node.check(node) &&
-        n.SourceLocation.check(node.loc)) {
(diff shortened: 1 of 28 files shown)


eventualbuddha:
Totally fair if you'd rather not do this, @benjamn.

benjamn:
I'm not totally against it, but maybe we can do it more incrementally? Like, if you're editing a particular file, and reformatting it with prettier first makes the editing easier, then go for it.

I would be happy to convert the remaining 4-spaces-per-tab modules to 2-spaces-per-tab, which is a nice side effect of this conversion. I've been doing that incrementally to preserve blame history.

Historical note: Prettier is based on a fork of Recast's pretty printer, so this is also indirectly Recast rewriting itself!

eventualbuddha:
> I'm not totally against it, but maybe we can do it more incrementally?

Cool, I'll think about whether there's a way to partially incorporate this.

PR #644, opened 2019-10-16
refactor: use arrow functions where appropriate

eventualbuddha:
This was done via codemod:

```
npm i @codemod/cli @resugar/codemod-functions-arrow
./node_modules/.bin/codemod -p @resugar/codemod-functions-arrow {lib,test}/*.ts
```

This only rewrites functions that have a single statement that returns a value.

### DIFF
--- a/test/visit.ts
+++ b/test/visit.ts
@@ -26,9 +26,7 @@ describe("types.visit", function() {
         assert.ok(thisExp.test(withThis));
 
         types.visit(ast, {
-            visitThisExpression: function() {
-                return builders.identifier("self");
-            }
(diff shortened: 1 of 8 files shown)


test/visit.ts
```diff
@@ -26,9 +26,7 @@ describe("types.visit", function() {
         assert.ok(thisExp.test(withThis));
 
         types.visit(ast, {
-            visitThisExpression: function() {
-                return builders.identifier("self");
-            }
+            visitThisExpression: () => builders.identifier("self")
```
eventualbuddha: Hmm. These probably shouldn't be arrow functions in case someone wants to edit them to reference `this`.

PR #676, opened 2019-12-06
feat: recast:esprima supports jsx

shuerguo999:
recast esprima supports jsx

### DIFF
--- a/parsers/esprima.ts
+++ b/parsers/esprima.ts
@@ -18,7 +18,8 @@ export function parse(source: string, options?: any) {
     onComment: comments,
     range: getOption(options, "range", false),
     tolerant: getOption(options, "tolerant", true),
-    tokens: true
+    tokens: true,
+    jsx: getOption(options, "jsx", false)
   });
 
   if (!Array.isArray(ast.comments)) {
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Thanks! Could you add a test that would fail without this change?

shuerguo999:
Ok， here is a demo:

`const recast = require('recast');
const code = '
    class Square extends React.Component {
        render() {
            return (
                <button className="square">
                    {this.props.value}
                </button>
            );
        }
    }';
const ast = recast.parse(code);`

Error message:

> throw this.unexpectedTokenError(token, message);
                ^
Error: Line 5: Unexpected token <
    at ErrorHandler.constructError

eventualbuddha:
> Ok, here is a demo.

Thanks, could you add it to the branch as a test?

eventualbuddha:
I ended up adding esprima to the existing JSX tests that were only exercising babel. They all passed except for the last one, which illustrated that [esprima does not support JSX fragments](https://github.com/jquery/esprima/issues/2020). I disabled that test for esprima, but it's likely to be an issue for anyone attempting to use esprima with JSX.

eventualbuddha:
Released in v0.19.0

PR #688, opened 2020-02-08
Fix Flow function type param

moroine:
Fix issues around:
- FunctionTypeParam when used without parenthesis like `number => string`.
- TypeAlias of FunctionTypeAnnotation


eventualbuddha:
Published in v0.18.6.

PR #694, opened 2020-03-22
add parentheses for function expressions for TSAsExpression

zxbodya:
Ensure parentheses are printed for the case:

```ts
const highlight = (n => false) as (a: number) => boolean;
```


eventualbuddha:
Released in v0.18.10

PR #695, opened 2020-03-22
babel 7.9, support for typescript 3.8

zxbodya:
Updating `printExportDeclaration` to expect `exportKind` to be not null for things like `export interface`


eventualbuddha:
What’s an example of something that requires this? Can you add tests?

zxbodya:
I am using recast in https://github.com/zxbodya/flowts which basically converts Flow typed sources to TypeScript. 

Currently, I am working on adding support for `export type / import type`  syntax introduced in typescript 3.8, which requires using babel 7.9.

About tests - it is already covered (the existing tests will fail with a new babel version).

The current recast version when using with a new babel will add extra `type` in many cases where it is not expected, like `export type interface {…}` or `export type type A = …`

PR #699, opened 2020-04-05
fix precedence for ** operator, to correctly parenthesize expressions

zxbodya:
Input code:

```
2 * (n / 2) ** 0.5;
```

Was printed as:

```
2 * (n / 2 ** 0.5);
```

After the fix would be printed as in input.

was discovered here: https://github.com/zxbodya/flowts/issues/4


eventualbuddha:
Thanks!

eventualbuddha:
Released in v0.18.9

PR #701, opened 2020-04-07
feat(Printer): add wrapPrint config option

jedwards1211:
I want to print out ASTs with certain nodes highlighted, adding this option
to wrap `print`/`printGenerically` would allow me to do that since they're
called recursively for every node.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -78,7 +78,7 @@ const Printer = function Printer(this: PrinterType, config?: any) {
         return (path: any) => print(path, options);
     }
 
-    function print(path: any, options: any) {
+    let print = function print(path: any, options: any) {
         assert.ok(path instanceof FastPath);
         options = options || {};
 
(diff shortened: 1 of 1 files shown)


benjamn:
Thanks for the suggestion, but I don't think this is the right level of abstraction to be adding a do-anything hook like this. It would mean exposing `Lines` objects to user code, which are currently just an implementation detail of the Recast printer. What sort of highlighting did you have in mind, exactly?

jedwards1211:
Sometimes in debugging `jscodeshift` finding/filtering it would be nice to be able to print out the code with the nodes in a collection highlighted (using `chalk` for example)

jedwards1211:
Yeah I guess the approach I just outlined would work for me, not as direct as I was hoping but as long as there's no obscure way a printed and reparsed AST can differ in structure, it will work

jedwards1211:
Okay yup, I got it to work without this PR: https://github.com/codemodsquad/print-highlighted-ast

PR #708, opened 2020-04-26
Update build dependencies

eventualbuddha:
Some can't be updated yet, like mocha, which requires nodejs 8+.


benjamn:
@eventualbuddha The `eslint` source code uses no-binding `} catch {` syntax, which effectively excludes Node.js 8, so I think we can stop worrying about running tests in Node.js 8 or earlier, now that we're running `eslint` as part of the tests. Recast should still (probably?) work in earlier versions of Node, but they won't be tested.

PR #709, opened 2020-04-27
fix: account for invalid shorthand properties

### LINKED ISSUE #704: Ignores differing ObjectProperty key/value whenever shorthand: true
Obviously it's a programmer mistake to have `shorthand: true` whenever the key/value differ.  But let me make a case for why it's probably good to compare the key/value when printing anyway...

I have to do my codemod with @babel/traverse to guarantee that certain renamings respect `let`, `const`, `var`, etc scope.

In a test I was seeing `@babel/generate` output `{value: value0}` in one destructuring, but `recast.print` was outputting just `{value}`.  Looking at the AST, I saw that it was because `path.scope.rename` fails to clear the `shorthand` flag.

Seems like it would be worth comparing the `key` and `value` even on shorthand properties, just in case?

```
                      {
                        "type": "ObjectProperty",
                        "start": 56,
                        "end": 61,
                        "loc": {
                          "start": {
                            "line": 4,
                            "column": 20
                          },
                          "end": {
                            "line": 4,
                            "column": 25
                          }
                        },
                        "method": false,
                        "key": {
                          "type": "Identifier",
                          "start": 56,
                          "end": 61,
                          "loc": {
                            "start": {
                              "line": 4,
                              "column": 20
                            },
                            "end": {
                              "line": 4,
                              "column": 25
                            },
                            "identifierName": "value"
                          },
                          "name": "value"
                        },
                        "computed": false,
                        "shorthand": true,
                        "value": {
                          "type": "Identifier",
                          "start": 56,
                          "end": 61,
                          "loc": {
                            "start": {
                              "line": 4,
                              "column": 20
                            },
                            "end": {
                              "line": 4,
                              "column": 25
                            },
                            "identifierName": "value"
                          },
                          "name": "value0"
                        },
                        "extra": {
                          "shorthand": true
                        }
                      }
```

eventualbuddha:
Closes #704

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -774,7 +774,7 @@ function genericPrintNoParens(path: any, options: any, print: any) {
             parts.push(key);
         }
 
-        if (! n.shorthand) {
+        if (!n.shorthand || n.key.name !== n.value.name) {
             parts.push(": ", path.call(print, "value"));
         }
 
--- a/test/printer.ts
+++ b/test/printer.ts
@@ -2163,4 +2163,31 @@ describe("printer", function() {
     const pretty = new Printer().printGenerically(ast).code;
     assert.strictEqual(pretty, code);
   });
+
+  it("does not print shorthand properties if key != value", function () {
+    const code = [
+      "var o = {",
+      "    value: value0",
+      "};"
+    ].join(eol);
+
+    const ast = b.program([
+      b.variableDeclaration("var", [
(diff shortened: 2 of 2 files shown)


jedwards1211:
Yeah!  It looks like that code change was pretty simple, I'll try to PR fixes for stuff like this in the future.

eventualbuddha:
Published in v0.19.1.

PR #712, opened 2020-05-25
Prettier, linting, and significant improvements to parens

### LINKED ISSUE #589: Formatting of await misses parentheses
The following two functions differ, in that one awaits the result of `a()` while one dereferences `b` off of `a`'s return value and then awaits that value.

```js
async function fnA() {
  return await a().b;
}
async function fnB() {
  return (await a()).b;
}
```

The AST for the latter produces the former in `recast`:

```js
const {builders} = require('ast-types');

console.log(
  recast.print(
    builders.functionDeclaration(
      builders.identifier('fnB'),
      [],
      builders.blockStatement([
        builders.returnStatement(
          builders.memberExpression(
            builders.awaitExpression(builders.callExpression(builders.identifier('a'), [])),
            builders.identifier('b')
          )
        ),
      ])
    )
  ).code
);

// produces

function fnB() {
    return await a().b;
}
```

Moreover, I can't find a workaround for this bug - I'd like to explicitly wrap the `await` expression in parentheses but don't see a way to do so. **EDIT:** I missed the `ParenthesizedExpression` type.

### LINKED ISSUE #705: Parenthesis on return statement has been removed
Reproduce example using jscodeshift:

https://astexplorer.net/#/gist/177972c375c147cd128f46354af81057/ae5a5014fa96340436dd3c68ff5063767dea8f4a

A simple demo without JSX
https://astexplorer.net/#/gist/ee27a86037e45714a0e4acaa57eca5c2/1a769021f57b29d21b3de0d8b59b44236decbcff

```
class Demo extends PureComponent {
   render() {
    return (
      <>
        <View
          style={{ backgroundColor: color.black }}
        />
        <View />
      </>
    );
  }
}

export default Demo;
```
to
```
class Demo extends PureComponent {
   render() {
    return <>
      <View
        style={{ backgroundColor: color.white }}
      />
      <View />
    </>;
  }
}

export default Demo;

```

I just want to change `color.black` to `color.white`, but parens has been removed.

### LINKED ISSUE #711: (await a).b becomes await a.b

### LINKED ISSUE #715: Typescript: [...any[]] becomes [...any[][]]

conartist6:
Fixes #715
Fixes #705
Fixes #589


lib/fast-path.ts
```diff
@@ -305,52 +305,43 @@ FPp.needsParens = function(assumeExpressionContext) {
...
-    return false;
-  }
-
-  switch (node.type) {
   case "UnaryExpression":
   case "SpreadElement":
   case "SpreadProperty":
-    return parent.type === "MemberExpression"
-      && name === "object"
-      && parent.object === node;
```
conartist6, line 337: Both of these checks do exactly the same thing. I eliminated the duplication and replaced all the name checks with reference equality checks against node.

lib/printer.ts
```diff
@@ -2507,7 +2494,6 @@ function genericPrintNoParens(path: any, options: any, print: any) {
     case "XMLComment":
     case "XMLProcessingInstruction":
     default:
-        debugger;
```
conartist6, line 2510: Not sure if there was any particular reason this was here. I found it when I went to make sure I didn't leave any of my own debuggers in.

lib/printer.ts
```diff
@@ -2147,17 +2147,13 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         ]);
 
     case "TSAsExpression": {
-        var withParens = n.extra && n.extra.parenthesized === true;
-        if (withParens) parts.push("(");
         const expression = path.call(print, "expression");
-        const expressionType = path.getValue().expression.type;
-        const needParens = expressionType === "ArrowFunctionExpression" || expressionType === "FunctionExpression";
```
conartist6, line 2154: My testing indicates that `FunctionExpressions` do not need parens when they are the `expression` of a `TSAsExpression`. That is to say, `(function() {} as () => number)` is parseable. I also verified the failing case: `(() => {} as () => number)` is not parseable, and requires the arrow function to be wrapped in parens.

conartist6, line 2154: `FunctionExpression` was not needed here. See new TS parens tests

lib/fast-path.ts
```diff
@@ -438,60 +433,59 @@ FPp.needsParens = function(assumeExpressionContext) {
...
-        n.ExportDefaultDeclaration.check(parent) &&
-        n.FunctionExpression.check(node.callee)) {
+    if (
+      parent.declaration === node &&
+      n.ExportDefaultDeclaration.check(parent) &&
+      n.FunctionExpression.check(node.callee)
+    ) {
       return true;
     }
+    break;
```
conartist6, line 485: There really aught to be lint rules to catch this sort of thing. It's a potential bug if not an actual one.

lib/fast-path.ts
```diff
@@ -393,18 +383,24 @@ FPp.needsParens = function(assumeExpressionContext) {
...
+    case "LogicalExpression":
+      return true;
+
     case "CallExpression":
-    case "MemberExpression":
     case "NewExpression":
+      return parent.callee === node;
+
     case "ConditionalExpression":
-    case "YieldExpression":
```
conartist6, line 406: Unnecessary. `yield yield true` parses correctly.

conartist6:
I wonder if there is a way to make `needsParens` more generic, or at very least less combinatoric. The work with operators seems like a good starting point.

lib/fast-path.ts
```diff
@@ -384,48 +374,31 @@ FPp.needsParens = function(assumeExpressionContext) {
...
-    switch (parent.type) {
-    case "BinaryExpression":
-    case "LogicalExpression":
-    case "UnaryExpression":
-    case "SpreadElement":
-    case "SpreadProperty":
-    case "CallExpression":
-    case "MemberExpression":
-    case "NewExpression":
-    case "ConditionalExpression":
```
conartist6, line 405: These only require parens in certain positions. Once this and `YieldExpression` are accounted for it is clear that `yield` and `await` behave the same was as assignment and conditional expressions.

conartist6:
I've split this into many focused commits. @benjamn if you get around to reviewing this hopefully it will be easier to see what's going on commit by commit.

eventualbuddha:
For what it’s worth,  I tried to do a Prettier-reformat in https://github.com/benjamn/recast/pull/643. @benjamn is of course welcome to change his mind, but it may be that he’d prefer to leave it as-is. I’m happy to review the PR, but my default assumption, if @benjamn doesn’t say otherwise, is that we should not reformat for the sake of it.

conartist6:
The biggest pain about running prettier everywhere is just the conflicts it creates with other people's PRs.

I don't personally agree with @benjamn that prettier changes too much at once. Yes, devs will have to skip past the formatting commit to see other history, but most devs know how to do that. Formatting incrementally is worse in my mind because then individuals making changes need to format a lot of code to change a little. And you don't want them to mix formatting and logic changes because that makes it much harder to see the actual logic changes. So then you're essentially requiring that all contributors know how to correctly handle formatting your library (first make a formatting commit, then your changes), which is less realistic than just setting up formatting so your contribs don't have to worry about how their code should be quoted, indented, etc.

conartist6:
It gets nasty too if there's a prettier config (needed for incremental reformatting) but not all code follows it. Now you have to bug people who make unformatted changes to formatted files. Also if someone makes extensive changes and then formatting is triggered they may now face the unenviable task of trying to go back and try to "unformat" to separate their logic changes from formatting changes so they can be separate commits. I know I've had to do that before. No fun.

eventualbuddha:
I agree with you, but I don’t make the call 🤷🏻‍♂️. If you remove the formatting commit, I’ll review and hopefully merge & release. Or you can make your case to @benjamn here. Perhaps do both?

conartist6:
It's going to be quite a bit of work to unrebase all my changes from the prettier commit. I know because I did a lot of it going the other way to introduce the prettier commit... So I guess I'll wait a little longer to see if @benjamn shows up. I already released a forked package from this PR under my personal npm namespace so I'm unblocked.

benjamn:
Hoping to get to this this weekend. Generally ok with Prettier if we think it will encourage/ease contributions to the library. I tend to fight with it, but I'm not actively working on this code that much anymore, so my preferences should be taken with a grain of salt.

Fun fact: Prettier began as a fork of the Recast pretty printer! If you check out their version of [`needsParens`](https://github.com/prettier/prettier/blob/master/src/language-js/needs-parens.js) you may experience a strong sense of déjà vu.

@conartist6 To your question about making `needsParens` more generic, I do believe we might be able to move the precedence rules into https://github.com/benjamn/ast-types, so that we wouldn't need a big `switch` statement, at least for expressions. I will happily tag you when I get around to that refactoring, but this PR doesn't need to wait for that to happen.

conartist6:
Wow that is deja vu given how much time I spent looking at that file. I had the chance to talk with Christopher Chedeau a couple of times as well. Thanks for taking a look!

conartist6:
@benjamn ping

skeggse:
I'm on-board with Prettier! Should I close #349 in favor of this, then?

lib/fast-path.ts
benjamn: They don't exactly always mean the same thing. You could have a node that's reused for more than one child, so `parent.property === node`, but the name of the child along the current traversal path is not actually `property`, because we're printing a different child right now. This is an obscure point for sure, but I'd like to keep the existing (perhaps excessively careful) logic.

lib/printer.ts
benjamn: This is a common place where printing can crash, because an unimplemented type is encountered. I've found it helpful to have a `debugger` statement already here, so that I can quickly debug these cases.

lib/fast-path.ts
```diff
@@ -305,238 +322,170 @@ FPp.needsParens = function(assumeExpressionContext) {
...
   // itself, then it doesn't need parentheses. Only Node objects (in fact,
   // only Expression nodes) need parentheses.
   if (this.getValue() !== node) {
     return false;
   }
 
-  // Only statements don't need parentheses.
-  if (n.Statement.check(node)) {
-    return false;
-  }
```
benjamn, line 320: This style of check is important, since `Statement` is an abstract supertype with many concrete subtypes. Just checking `node.type === "Statement"` is not the same. I'll put this back, but just wanted you to be aware of the difference.

benjamn:
@conartist6 @eventualbuddha Ok, I finally got around to this—sorry for the _loooong_ wait!

Regarding the Prettier reformatting, I found there were a lot fewer trivial changes if I kept my own personal style of using double quotes for strings, which I'm happy to let Prettier enforce.

I removed Husky because I don't like fighting with pre-commit hooks, but eslint now runs before tests, so we won't forget to run it. I realize this means Prettier won't automatically run whenever contributors create commits, but we can easily ask them to run `npm run format` before merging their PRs.

In case I messed up anything in the process of rebasing/adjusting these commits, your original branch is preserved as [`conartist6-parens-fixes-original`](https://github.com/benjamn/recast/tree/conartist6-parens-fixes-original), for comparison.

Thanks for doing all this work, and thanks for continuing to ping me to review it. It would have happened sooner without the Prettier changes, but I don't mind having that incorporated into the repository now.

lib/fast-path.ts
conartist6: Aaaaaah OK right. Thanks!

benjamn:
Just published these changes in [`recast@0.20.0`](https://www.npmjs.com/package/recast/v/0.20.0).

conartist6:
@benjamn Thanks so much. Seeing progress restores a little of the faith I lost after discovering just how many popular projects aren't being maintained at all.

PR #716, opened 2020-06-12
Fix trailing comma in destructuring rest spread

### LINKED ISSUE #559: Trailing comma in destructuring rest spread property breaks code
When `print` option `trailingComma` used, and one of nodes in object destructuring removed, trailing comma added to spread operator which leads to broken code:

```js
const { ...b,} = obj
           ^

SyntaxError: Rest element must be last element
```

Code to reproduce:

```js

const source = `
    const {
        a,
        ...b
    } = obj;
`;

    parser: {
        parse: babel.parse,
    }
});

let once = false;
traverse(babelAST, {
    enter(path) {
        if (once)
            return;

        if (path.type !== 'ObjectProperty')
            return;

        path.remove();
        once = true;
    }
});

console.log(recast.print(babelAST, {
    trailingComma: true,
}));
// outputs
// PrintResult { code: '\n    const {\n        ...b,\n    } = obj;\n' }
```
Will be great to not add comma when rest spread used :)

Daniel15:
When the `trailingComma` option is enabled, we shouldn't append a trailing comma to a rest spread.

Closes #559

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -698,7 +698,11 @@ function genericPrintNoParens(path: any, options: any, print: any) {
             allowBreak = !multiLine;
           } else if (len !== 1 && isTypeAnnotation) {
             parts.push(separator);
-          } else if (!oneLine && util.isTrailingCommaEnabled(options, "objects")) {
+          } else if (
+            !oneLine &&
+            util.isTrailingCommaEnabled(options, "objects") &&
+            childPath.getValue().type !== "RestElement"
+          ) {
             parts.push(separator);
           }
           i++;
--- a/test/printer.ts
+++ b/test/printer.ts
@@ -836,6 +836,28 @@ describe("printer", function () {
     },
   );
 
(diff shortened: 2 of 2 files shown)


lib/printer.ts
```diff
@@ -717,7 +718,11 @@ function genericPrintNoParens(path: any, options: any, print: any) {
                     allowBreak = !multiLine;
                 } else if (len !== 1 && isTypeAnnotation) {
                     parts.push(separator);
-                } else if (!oneLine && util.isTrailingCommaEnabled(options, "objects")) {
+                } else if (
+                    !oneLine && 
+                    util.isTrailingCommaEnabled(options, "objects") &&
+                    childType !== 'RestElement'
```
eventualbuddha, line 724: Inline `childType`? Unless I’m missing why it should be hoisted.

Daniel15: Will do!

Daniel15:
Thanks for merging!

PR #726, opened 2020-08-23
Use tslib to avoid repeatedly inlining helper definitions.

benjamn:
Much like [`@babel/runtime`](https://www.npmjs.com/package/@babel/runtime) for Babel, the [`tslib`](https://www.npmjs.com/package/tslib) package is a runtime library that TypeScript can leverage to emit fewer helper function declarations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3232,6 +3232,14 @@
       "dev": true,
       "requires": {
         "tslib": "^1.9.0"
+      },
+      "dependencies": {
+        "tslib": {
+          "version": "1.13.0",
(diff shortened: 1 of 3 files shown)


benjamn:
@eventualbuddha Any concerns about this? The 2.0 release of `tslib` was recent, so there may be lots of applications still using 1.x, but I think it's probably fine if `recast` ends up with its own nested `node_modules/tslib` directory?

eventualbuddha:
I don’t have any particular insight into this question, but I also don’t see any reason this would be bad in a noticeable way for users of recast, so go for it!

PR #744, opened 2020-08-30
fix: parenthesizes around arrow function body returning object

### LINKED ISSUE #743: Brocken parenthesizes around arrow function body when return an object
`recast` produces invalid code when `babel` used as a parser for such code, that returns an object:

before:
```js
 const get = (a) => ({
    hello: 'world',
    'x': 'y',
});

```
after:
```js
 const get = (a) => {
    hello: 'world',
    'x': 'y',
};
```
Which is invalid because has different meaning without parenthesizes, as a labeled series expressions without an end.

Can be reproduced using `v0.20.2` with:
```js

const source = `
    const get = (a) => ({
        hello: 'world',
        'x': 'y',
    });
`;

    parser: {
        parse: babel.parse,
    }
});

traverse(babelAST, {
    Identifier(path) {
        if (path.node.name === 'a')
            path.remove();
    }
});

console.log('recast:', recast.print(babelAST));
console.log('babel', generate(babelAST));
```

Output:

```js
recast: PrintResult {
  code: "\n    const get = () => {\n        hello: 'world',\n        'x': 'y',\n    };\n"
}
babel {
  code: "const get = () => ({\n  hello: 'world',\n  'x': 'y'\n});",
  map: null,
  rawMappings: undefined
}
```

Can  be related to https://github.com/benjamn/recast/issues/705
This can be fixed setting [createParenthesizedExpressions](https://babeljs.io/docs/en/babel-parser):

```js
    parser: {
        parse: () => {
            return babel.parse(source, {
                createParenthesizedExpressions: true
            });
        }
    }
});
```

But would be amazing if this option can be avoided, because it makes harder to [traverse](https://github.com/coderaiser/putout/tree/master/packages/traverse) and [compare](https://github.com/coderaiser/putout/tree/master/packages/compare) `AST` nodes, because `ParenthesizedExpressions` adds an extra noise.

coderaiser:
Fixes #743. Using `recast` with `babel-parser` and `traverser` break code with arrow function expression returning an object.

Code:

```js
(a) => ({a: 'b'});
```

Becomes:

```js
() => {a: 'b'}
```
After removing function argument from `@babel/traverse`.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -391,7 +391,11 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         );
       }
 
-      parts.push(" => ", path.call(print, "body"));
+      if (n.body.type === 'ObjectExpression') {
+          parts.push(" => ", '(', path.call(print, "body"), ')');
+      } else {
+          parts.push(" => ", path.call(print, "body"));
+      }
 
       return concat(parts);
 
--- a/test/printer.ts
+++ b/test/printer.ts
@@ -1031,6 +1031,25 @@ describe("printer", function () {
     const pretty = printer.printGenerically(ast).code;
     assert.strictEqual(pretty, code);
   });
+  
+  it("adds parenthesis around arrow functions body when returning object expression using babel parser", function () {
+    const expected = "() => ({a: 'b'});";
+    const source = "(a) => ({a: 'b'});";
+    const ast = recast.parse(source, {
+        parser: require('@babel/parser'),
+    });
+    const traverse = require('@babel/traverse').default;
+    
+    traverse(ast, {
+        Function(path: any) {
+            path.get('params.0').remove();
+        }
+    });
+
+    const printer = new Printer();
+    const result = printer.print(ast).code;
+    assert.strictEqual(result, expected);
+  });
 
   it("prints class property initializers with type annotations correctly", function () {
     const code = ["class A {", "  foo = (a: b): void => {};", "}"].join(eol);


eventualbuddha:
This seems okay at first glance, but then I don’t understand why it isn’t already handled by this: https://github.com/benjamn/recast/blob/8cc1f42408c41b5616d82574f5552c2da3e11cf7/lib/fast-path.ts#L500-L507.

coderaiser:
> I don’t understand why it isn’t already handled by this

Looks like it doesn't handled by `ObjectExpression` `switch case` from `fast-path`, because it is:
- not get into`ObjectExpression` switch in the `printer`, it get only to `ArrowFunctionExpression`
- so if `fast-path` `ArrowFunctionExpression` `switch-case` will be updated, result will be something like this:
```js
(b => {a: 'b'})
```
- `fast-path` stack doesn't have `ObjectExpression`, it ends on `ArrowFunctionExpression`
Screen Shot 2020-08-31 at 11 39 48 AM

coderaiser:
Looks like this code https://github.com/benjamn/recast/blob/v0.20.3/lib/patcher.ts#L478-L494 returns `false` and exits loop when it checks `params`, so there is no recursive execution for `body` which is `ObjectExpression` and no check for parenthesis.

coderaiser:
any progress on this?

coderaiser:
Would be amazing to merge this, or fix in any other way, so I don't have to support my own fork of [recast](https://github.com/coderaiser/recast) accessible by `npm i @putout/recast`, where this bug is fixed already :).

zxbodya:
Hi,
caught the same issue, and probably found the change causing it - made a pr reverting that change and adding a test from this PR - https://github.com/benjamn/recast/pull/830

conartist6:
Yep, it was me that broke this, and I also have a [duplicate issue](https://github.com/benjamn/recast/issues/1067) and [a good fix](https://github.com/benjamn/recast/pull/1068). I'm hoping to get the attention of @benjamn to resolve this, and if not I'm actually more than happy to maintain a fork so one way or another I think we can move forward on this finally!

conartist6:
I'm also working full time on tooling that will make it significantly easier to develop packages in a decentralized way, that is to say without the single point of failure of maintainers who are the sole holders of publish perms to the central npm repo and whose focus tends to shift for perfectly natural and understandable reasons.

gnprice:
The issue appears to have been fixed now by #1068, so I believe this PR can be closed.

eventualbuddha:
Verified this no longer reproduces. Thanks for the nudge, @gnprice.

PR #766, opened 2020-09-17
Bumped ast-types version

ethancrook99:


ethancrook99:
I would really appreciate if you could get this merged quickly, it would help me out a lot!

ethancrook99:
I'm not sure how to fix the failing test... I looked through the code, but I couldn't identify what change I needed to make to handle the new types. Any pointers?

eventualbuddha:
My guess is that the new ast-types adds a new node type that recast doesn't know about?

ethancrook99:
@eventualbuddha It's the new ChainElement and ChainExpression types, I'm just not sure how to make recast recognize new types

mjfaga:
I updated the test to work past this here: https://github.com/benjamn/recast/pull/769

benjamn:
@ethancrook99 The reason these tests exist is to ensure that the Recast pretty-printer is able to _print_ the new AST types, so the conservative pretty-printing algorithm can fall back to generating valid (if somewhat reformatted) code wherever the original AST has been modified.

Where the AST has _not_ been modified, Recast is able to copy the original code directly to the output (give or take some reindentation). In other words, Recast can often get away with not actually knowing how to print unchanged AST nodes, but it will fail badly when AST nodes involving a `ChainExpression` are modified or added to the AST, because it can't just copy the original code in those cases.

If you need `ast-types@0.14.2` specifically because you want to write AST transforms using Recast that involve `ChainExpression` AST nodes, then you will almost certainly run into cases where you need Recast to be able to print these nodes, so I trust you can understand why simply adding the new types to the `switch` statement without actually implementing them (like #769 does) would be side-stepping the issue.

In short, the test failures are telling us that we need to add printer cases to [`lib/printer.ts`](https://github.com/benjamn/recast/blob/19f0bcc0171d538974f490ab63aa82a13e21ca06/lib/printer.ts#L244) (where you can find lots of examples of similar code).

ethancrook99:
@benjamn I can give it a shot! I'll let you know if I need any guidance

ethancrook99:
I added possible implementations for the new types in `printer.ts`, but I feel like I may have oversimplified `ChainExpression`. I just copied directly from `ExpressionStatement`--should I have done something different here?

lib/printer.ts
```diff
@@ -308,6 +308,20 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       return concat(parts);
     }
 
+    case "ChainExpression":
+      return concat([path.call(print, "expression"), ";"]);
```
eventualbuddha, line 312: I believe this will insert a semicolon at the end of a genetically printed chain expression, which is wrong. Perhaps just use the value of `path.call(print, "expression")`? You could also add a test where adding a semicolon would fail, such as here: `a(b?.c)`.

eventualbuddha (changes requested):
Close, should be easy to fix.

lib/printer.ts
ethancrook99: I made that change, thanks for the review! I tried to make a test, but I couldn't quite figure out how to get one working. Here's what I have, where am I going wrong?

```
  it("does not print semicolons for ChainExpression", function () {
    const node = b.chainExpression(
      b.expression("a(b?.c)"),
    );

    assert.strictEqual(recast.print(node).code, "a(b?.c)");
  });
```

mjfaga:
Thanks for putting in the remaining effort on this. I hadn't had a chance to review the patterns in play just yet and noticed the `TODO`s in there and followed that in the short term to resolve an issue I ran into locally on this very thing which did not require printing the Optional Chain operator. 

Apologies for the side-step @benjamn 👍

eventualbuddha:
Wasn't there a test? Where'd that go?

ethancrook99:
I never actually added the test, I couldn't figure out how to get it working properly. I put the code that I did have in my reply to the above thread though, if you have any suggestions on how to fix it then I'll add the test back in

ethancrook99:
Can I get some guidance on how to make a test for `"a(b?.c)"` not printing semicolons? I can't quite figure out how to follow the same pattern as the other printer tests.

eventualbuddha:
I wouldn't worry about the semicolon thing specifically. Just have a test that generically prints a chain element.

ethancrook99:
@eventualbuddha I added a couple of tests, looks like everything is working together nicely now

eventualbuddha (approved):

benjamn:
@ethancrook99 @eventualbuddha Just published these changes to npm as `recast@0.20.4`, with some additional tests and tweaks in #783. Thanks for working on this! 🙌

PR #783, opened 2020-10-12
Add more tests of chain expressions.

benjamn:
Follow-up to #766, addressing a few issues:

1. The `ChainElement` type is an abstract supertype of `CallExpression` and `MemberExpression` (see https://github.com/benjamn/ast-types/pull/399), so [`switch (n.type)`](https://github.com/benjamn/recast/blob/ae0a51e9e75b892e8d6aa26eef6e6976742aab9a/lib/printer.ts#L244) will always see either `"CallExpression"` or `"MemberExpression"` and never `"ChainElement"`.

2. We need to keep the printing of `{Member,Call}Expression` aligned with the printing of `Optional{Member,Call}Expression`, at least until the latter types are fully deprecated by all parsers.

3. The optionality logic for call expressions was slightly off. Whether a `?.` is printed should not depend on the optionality of `n.callee`. This has been an issue since #511, per git blame.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -297,7 +297,10 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       parts.push(path.call(print, "object"));
 
       const property = path.call(print, "property");
-      const optional = n.type === "OptionalMemberExpression" && n.optional;
+
+      // Like n.optional, except with defaults applied, so optional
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):

benjamn:
@eventualbuddha Thanks for the review!

PR #804, opened 2020-11-10
feature: add import assertions support

coderaiser:
According to stage 3 proposal https://github.com/tc39/proposal-import-assertions.


lib/printer.ts
```diff
@@ -602,7 +602,14 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         parts.push(" from ");
       }
 
-      parts.push(path.call(print, "source"), ";");
+      parts.push(path.call(print, "source"));
+
+      if (n.assertions && n.assertions.length > 0) {
+          const [assert] = n.assertions;
```
eventualbuddha, line 608: I see from the proposal that there's only one key they expect to be in use for now, `type`. But is there any reason not to print however many assertions as there are in the AST?

coderaiser: There is no reason no to print as many assertions as there are in the AST. Just added.

test/babel.ts
```diff
@@ -191,6 +191,29 @@ describe("Babel", function () {
     );
   });
 
+  it("ImportAttribute", function () {
+    const code = [
+      `import foo from 'foo';`,
```
eventualbuddha, line 196: Can you parse it too? Or do you have to build the assertions programmatically?

lib/printer.ts
```diff
@@ -602,7 +602,15 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         parts.push(" from ");
       }
 
-      parts.push(path.call(print, "source"), ";");
+      parts.push(path.call(print, "source"));
+
+      if (n.assertions && n.assertions.length > 0) {
+          for (const assert of n.assertions) {
+              parts.push(' assert', ' {', assert.key.name, ': ', `'${assert.value.value}'`, '}');
```
eventualbuddha, line 609: This would generate e.g. ` assert { type: 'json' } assert { foo: 'bar' }` given multiple assertions. Reading through the [parser](https://github.com/babel/babel/blob/5bbad8936b05a2b6739dcade6464dbc361118aa9/packages/babel-parser/src/parser/statement.js#L2192-L2247), it should be ` assert { type: 'json', foo: 'bar' }`. Also, I don't think we should be handling printing `Identifier` and `StringLiteral` by unwrapping them here, right? We should let recursive `print` calls handle them.

test/babel.ts
coderaiser: Just pushed  3db34e8 with new tests.

lib/printer.ts
coderaiser: Switched to `print` in 3d2bccb.

eventualbuddha:
Looks like @benjamn did this something similar (much later) in https://github.com/benjamn/recast/commit/8f1bcee915000ce951e31a999d598ee9e59ca586.

PR #815, opened 2020-12-04
fix: TSTypeAnnotation printer

gus3inov:
Fixes: https://github.com/benjamn/recast/issues/814

Now instead of `type Message = (params: Params) : typeof add;` we are get `type Message = (params: Params) => typeof add;`

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2220,9 +2220,9 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         path.getParentNode(0),
       );
 
-      let prfx = isFunctionType ? "=> " : ": ";
+      const prefix = isFunctionType ? "=> " : ": ";
 
-      return concat([prfx, path.call(print, "typeAnnotation")]);
+      return concat([prefix, path.call(print, "typeAnnotation")]);
     }
 
     case "TSIndexSignature":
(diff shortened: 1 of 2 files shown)


lib/printer.ts
```diff
@@ -2215,8 +2215,15 @@ function genericPrintNoParens(path: any, options: any, print: any) {
...
       return concat([path.call(print, "expression"), "!"]);
 
-    case "TSTypeAnnotation":
-      return concat([": ", path.call(print, "typeAnnotation")]);
+    case "TSTypeAnnotation": {
+      const isFunctionType = namedTypes.TSFunctionType.check(
+        path.getParentNode(0),
+      );
+
+      let prfx = isFunctionType ? "=> " : ": ";
```
eventualbuddha, line 2223:
```suggestion
      const prefix = isFunctionType ? "=> " : ": ";
```

lib/printer.ts
```diff
@@ -2215,8 +2215,15 @@ function genericPrintNoParens(path: any, options: any, print: any) {
...
+
+      return concat([prfx, path.call(print, "typeAnnotation")]);
```
eventualbuddha, line 2225:
```suggestion
      return concat([prefix, path.call(print, "typeAnnotation")]);
```

eventualbuddha (changes requested):
Thanks! Can you add a test please?

eventualbuddha:
Apologies for the long delay. The test you added seems to pass even without the change to `printer.ts` after rebasing on `master`, so I'm assuming something between when this PR was opened and now fixed the underlying issue.

PR #881, opened 2021-03-15
add sourceType option for esprima

zeratul1:
according to [distinguishing-a-script-and-a-module](https://esprima.readthedocs.io/en/4.0/syntactic-analysis.html#distinguishing-a-script-and-a-module,),
we should pass `sourceType` option to the `esprima.parse` function.

we have `esprima.parseScript`, `esprima.parseModule` and a deprecated `esprima.parse`.
[[API] Explicitly distinguish parsing a module vs a script #1576](https://github.com/jquery/esprima/issues/1576)


eventualbuddha:
Fixed in v0.21.4.

PR #885, opened 2021-03-19
Refine Prettier configuration

thorn0:
- pin Prettier's version in `package.json` (see https://prettier.io/docs/en/install.html)
- make `npm run format` format more files, not only `lib` and `test`
- run reformat

Also I noticed lint-staged is installed and configured, but husky isn't installed. lint-staged doesn't work without husky. Should I remove lint-staged?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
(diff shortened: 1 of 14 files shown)


fisker:
https://github.com/benjamn/recast/pull/712/commits/bb58688f40f8df1981d0ae44249a6cdab22e16c5

eventualbuddha:
Replaced by #1094.

PR #921, opened 2021-05-14
Typescript parser to support JSX

danieldelcore:
The Typescript parser is currently unable to parse JSX elements. 

Repro: [Codesandbox](https://codesandbox.io/s/old-firefly-yqdfh?file=/src/index.js) 

This is because the `jsx` babel plugin has not been added to the babel plugin array. 

**note:** I'm not sure if we should be using the [isTSX](https://babeljs.io/docs/en/babel-plugin-transform-typescript#istsx) option instead/as well as the `jsx` plugin

Related: https://github.com/fkling/astexplorer/issues/460


vjpr:
See also: https://github.com/facebook/jscodeshift/pull/291

Issue #922, opened 2021-05-14
Missing parentheses when argument of UnaryExpression is LogicalExpression !(a && b)


coderaiser:
When using such code:

```js
const recast = require("recast");
const {unaryExpression} = require('@babel/types');
const babel = require('@babel/parser');

const source = '(options || !options.bidirectional) ? false : true;';

const ast = recast.parse(source, {
    parser: babel,
});

ast.program.body[0].expression = unaryExpression('!', ast.program.body[0].expression.test);

console.log(recast.print(ast).code);
// outputs
!options || !options.bidirectional
```

Should output:

```js
!(options || !options.bidirectional)
```

Reproduced in latest version of `@babel/parser` when used with `@babel/core@7.14.2`

gnprice:
This issue no longer reproduces :tada: 

It reproduces as of `recast@0.21.0`:
```
$ cd $(mktemp -d)
$ npm i recast@0.21.0 @babel/types @babel/parser
$ xclip -o >test.js  # (to copy the code in the description to test.js)
$ node test.js
!options || !options.bidirectional;
```

It no longer reproduces as of 0.21.1:
```
$ npm i recast@0.21.1
$ node test.js
!(options || !options.bidirectional);
```

From looking through what changed between those versions, I think #1068 is probably the change that fixed it.

Issue #867, opened 2021-02-23
Additional semicolon it TSInterfaceBody


coderaiser:
When `AST` with `TSInterfaceBody` changed, it produced additional semicolon, which is not valid code:

```js
const {print, parse} = require('recast');
const code = [
    "interface Hello {",
    "  'hello': any;",
    "  'hello': string;",
    "}",
].join(eol);
    
const ast = recast.parse(code, {
    parser: require('recast/parser/typescript'),
});
    
ast.program.body[0].body.body.pop();
recast.print(ast).code;
// returns
"interface Hello {",
"  'hello': any;;",
"}",
```

Also I want to mention that `recast.prettyPrint` not cover this case, so it's not tested with `print`.

PR #923, opened 2021-05-14
Fix of broken parenthesis when argument of UnaryExpression is LogicalExpression !(a && b)

coderaiser:
Fixes of #922 #867


gnprice:
Both of these issues have since been happily fixed. So I think this PR can be closed.

eventualbuddha:
I verified that the test added in this PR passes on`master` now, so I'm closing it.

PR #987, opened 2021-09-11
Update source map example to use less misleading file names

### LINKED ISSUE #76: sourceMapName question
When I run the following code:

``` javascript
var recast = require('recast'),
    fs = require('fs');

var source = fs.readFileSync('index.js');

var result = recast.print(recast.parse(source, {
  sourceFileName: "index.js"
}), {
  sourceMapName: "map.json"
});

console.log(result.code); // Resulting string of code.
console.log(result.map); // JSON source map.
/*
results.map is:
{ version: 3,
  file: 'map.json',
  sources: [ 'index.js' ],
  names: [],
...rest of output elided...
*/

```

`result.map` contains a `file` property equal to "map.json". Maybe I'm missing something in the [spec](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/edit#), but I thought this property was supposed to be the name of the generated output file that the source map is associated with, _not_ the name of the map file itself. Furthermore, should the name of the map file ever be used in the generation of the map contents?

Also, I notice in `result.code` that the source map annotation (ie `//# sourceMappingURL=<url>`) is not provided. Are clients of recast responsible for printing this to the final output?

anka-213:
Fixes #76. See discussion in that issue.

I chose `source.min.js` as a hopefully obvious target name, since that's the context where source maps are used most often. I guess `target.js` could also have been a good name too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ All you have to think about is how to manipulate the syntax tree, and Recast wil
 var result = recast.print(transform(recast.parse(source, {
   sourceFileName: "source.js"
 })), {
-  sourceMapName: "map.json"
+  sourceMapName: "source.min.js"
 });
     
 console.log(result.code); // Resulting string of code.


eventualbuddha:
Fixed in v0.21.4.

Issue #831, opened 2020-12-23
Critical assertion error generating source maps 


GianlucaGuarini:
I am working since days on a Riot.js compiler bug, this issue is driving me mad. I have noticed that the problem is coming from the recast.print method when it tries to generate the sourcemaps.

I was able to reproduce the issue with a few lines of code [here](https://runkit.com/gianlucaguarini/recast-bug) 

That's the code to reproduce it in case you want to try to copy and paste it directly in your editor:

```js
const recast = require('recast')
const types = recast.types
const builders = recast.types.builders
const scope = builders.identifier('scope')

const expression = `(




                                                 {
    complete: onComplete,
    translateX: 100
  })`

const ast = recast.parse(expression, {
  sourceFileName: 'test.js',

})

// replace the scope of the identifiers found
types.visit(ast, {
  visitIdentifier(path) {
    path.replace(builders.memberExpression(
      scope,
      path.node,
      false
    ))

    return false
  }
})

// crate a function 
const result = builders.functionExpression(
  null,
  [scope],
  builders.blockStatement([builders.returnStatement(
    ast.program.body[0].expression
  )])
)

// generate the output
const output = recast.print(result, {
  sourceMapName: `source.map`
})

return output

```

Any hint or help solving this issue is really appreciated as long as the sourcemap output will be correct :)

Thank you

eventualbuddha:
Replacing an identifier with a member expression is not always safe. You’ll be replacing the keys in your object with e.g. `scope.translateX`, which is not a valid key.

GianlucaGuarini:
@eventualbuddha thank you for your feedback. The `@riotjs/compiler` is much more complex than this demo. The code above is just needed to show the issue in recast that seems to be clearly related to the breaking spaces since the following demo works as expected: https://runkit.com/gianlucaguarini/5fe44ae13e9cdb001a56fc21

eventualbuddha:
Ah, you're right. I don't know what's causing the issue you're seeing, but it does appear to be a bug. Thanks for clarifying.

GianlucaGuarini:
@benjamn do you accept PRs? I might be interested solving this issue with a patch

eventualbuddha:
Yep, though it may not happen quickly.

GianlucaGuarini:
@eventualbuddha I have tried to keep the patch simple https://github.com/benjamn/recast/pull/989 but your help on it is highly appreciated.

GianlucaGuarini:
[...]

PR #989, opened 2021-10-03
WIP: Fix https://github.com/benjamn/recast/issues/831

GianlucaGuarini:
Removing the `skipSpaces` optimization from the `Lines.getSourceMap` method does fix https://github.com/benjamn/recast/issues/831. [This test](https://github.com/GianlucaGuarini/recast/blob/3d9f488229e38e85994ccd7d4aa413d04fbe9dd7/test/mapping.ts#L81-L85) should be manually updated and it would be great if you could let me know if my patch goes into the right direction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 *.d.ts
 # Except...
 !/types/**/*.d.ts
+.idea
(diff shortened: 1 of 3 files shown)


GianlucaGuarini:
Closing this PR since it's stale for more than 1y

PR #1003, opened 2021-11-18
Babel-like `retainLines` printer option

michaelfig:
This is a minimum-hack-only implementation of `options.retainLines`, in the hopes it will be useful someday to actually finish the feature.

It currently only works for top-level statements, not anything nested.


michaelfig:
Closing because this is stale.

PR #1010, opened 2021-12-02
Update `ast-types` dependency to version 0.15.2

benjamn:
As with most `ast-types` updates, there are some new AST node types that need to be supported by the Recast pretty-printer, for situations where the original source code cannot simply be preserved (for example, you're writing a Recast transform that modifies nodes of these types, or nearby nodes whose printing is entangled).


benjamn:
I'm going to go ahead and merge/publish this PR in a new minor version (0.21.0), but I will use the `next` tag on npm for now, so folks can do their own testing in the meantime (👋 @eventualbuddha and others), before the flood of Renovate/Dependabot update PRs starts.

PR #1013, opened 2021-12-10
Print class property definite assignment assertions

### LINKED ISSUE #1012: Cannot add typescript definite assignment assertions to class properties
Recast version: `0.20.5`
Node version: `14.18.2`
Mac OS X version: `12.0.1`

Repro: https://github.com/lokshunhung/recast-definite-repro

[AST explorer](https://astexplorer.net/#/gist/b7ac82659baa9eeb3612616764478269/1551dd3fe807088a8e37298c77643f48fef71987)

---

`recast` with `recast/parser/typescript` can remove [definite assignment assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#definite-assignment-assertions) for class properties by setting `ClassProperty.definite = false` (originally `true`)

```ts
// Before
class A {
  foo!: string;
}

// After
class A {
  foo: string;
}
```

but cannot add definite assignment assertions for class properties by setting `ClassProperty.definite = true` (originally `undefined`)

```ts
// Before
class A {
  foo: string;
}

// After
class A {
  foo: string;
}

// should be `foo!: string;`
```

lokshunhung:
Resolves #1012

This PR makes the printer aware of `definite` property in `ClassProperty` AST node, printing `!` when `true`

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1449,6 +1449,10 @@ function genericPrintNoParens(path: any, options: any, print: any) {
         parts.push("?");
       }
 
+      if (n.definite) {
+        parts.push("!");
+      }
+
       if (n.typeAnnotation) {
         parts.push(path.call(print, "typeAnnotation"));
       }
--- a/test/printer.ts
+++ b/test/printer.ts
@@ -1206,6 +1206,33 @@ describe("printer", function () {
     assert.strictEqual(pretty, code);
   });
 
+  it("prints 'definite' ClassProperty correctly", function () {
+    const code = ["class A {", "  foo!: string;", "}"].join(eol);
+
+    const ast = b.program([
+      b.classDeclaration(
+        b.identifier("A"),
+        b.classBody([
+          Object.assign(
+            b.classProperty(
+              b.identifier("foo"),
+              null,
+              b.typeAnnotation(b.stringTypeAnnotation()),
+            ),
+            { definite: true },
+          ),
+        ]),
+      ),
+    ]);
+
+    const printer = new Printer({
+      tabWidth: 2,
+    });
+
+    const pretty = printer.printGenerically(ast).code;
+    assert.strictEqual(pretty, code);
+  });
+
   it("prints static ClassProperty correctly", function () {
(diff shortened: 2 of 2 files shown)


eventualbuddha:
Thanks! Mind adding a test?

lokshunhung:
@eventualbuddha Sure, could you point me to the location where I should put [this](https://github.com/lokshunhung/recast-definite-repro/blob/9667f507474e5e39b9b77631b24bba6faaf4287b/src/index.test.js)?

eventualbuddha:
> @eventualbuddha Sure, could you point me to the location where I should put [this](https://github.com/lokshunhung/recast-definite-repro/blob/9667f507474e5e39b9b77631b24bba6faaf4287b/src/index.test.js)?

Maybe [here](https://github.com/benjamn/recast/blob/master/test/typescript.ts)?

lokshunhung:
@eventualbuddha I checked `test/typescript.ts` and it seems that it contains tests for checking syntax validity and reprinting

I put the tests inside `test/printer.ts` below [this](https://github.com/benjamn/recast/blob/99c4fb9fcf68d121d58d08cfc5f2ed974607470c/test/printer.ts#L1167)

lokshunhung:
I had to use `Object.assign(b.classProperty(...), { definite: true })` instead of `b.classProperty.from(...)` in the test, as it seems that `ast-types` does not handle the `definite` property in `ClassProperty` nodes 😅 

```ts
import * as assert from "assert";
import * as ast from "ast-types";

const b = ast.builders;

const n1 = Object.assign(
  b.classProperty(
    b.identifier("key"),
    null
  ),
  { definite: true }
);

const n2 = b.classProperty.from({
  key: b.identifier("key"),
  value: null,
  definite: true,
});

assert.deepStrictEqual(n1, n2);
```

```
Expected values to be strictly deep-equal: 
+ actual - expected ... Lines skipped 
 
  { 
    access: undefined, 
    comments: null, 
    computed: false, 
+   definite: true, 
    key: { 
... 
    value: null, 
    variance: null 
  }
```

PR #1057, opened 2022-02-04
Parenthesize function at start of sequence

djpohly:
The firstInStatement function was missing the first child of a
SequenceExpression because it needs to check both the type of the parent
*node* and the name of the parent *path element*, which are not the same
in this case.

This fix is a "minimal change" style; it should be correct even if it's
not beautiful.  A cleaner fix would probably involve changing the way
the loop finds nodes.

### DIFF
--- a/lib/fast-path.ts
+++ b/lib/fast-path.ts
@@ -643,9 +643,11 @@ FPp.firstInStatement = function () {
       return true;
     }
 
+    // s[i + 1] and s[i + 2] represent the array between the parent
+    // SequenceExpression node and its child nodes
     if (
       n.SequenceExpression.check(parent) &&
-      parentName === "expressions" &&
+      s[i + 1] === "expressions" &&
       childName === 0
     ) {
       assert.strictEqual(parent.expressions[0], child);
--- a/test/parens.ts
+++ b/test/parens.ts
@@ -338,6 +338,10 @@ describe("parens", function () {
     check("(function (){})()");
   });
 
+  it("should be added to function expressions at start of sequence", function () {
+    check("(function (){})(), 0");
+  });
+
   it("issues #504 and #512", function () {
     check("() => ({})['foo']");
     check("() => ({ foo: 123 }[foo] + 2) * 3");


eventualbuddha:
As far as I can tell, this seems safe. I'll merge it soon unless @benjamn objects.

coderaiser:
I got case that may be relevant, since it's about `SequenceExpressions`. For such code:

```js
const name = (FIRST || SECOND).VALUE || HELLO;
```

After instrumenting by coverage tool 🎩[ESCover](https://github.com/coderaiser/escover) I receive next results.

✅ `@babel/generate` produces correct result:

```js
const name = (__c4['🧨'](1, 13), ((__c4['🧨'](1, 14), FIRST) || (__c4['🧨'](1, 23), SECOND)).VALUE) || (__c4['🧨'](1, 40), HELLO);
  ```

❌ `recast` produces incorrect result:

```js
 const name = (__c4['🧨'](1, 13), (__c4['🧨'](1, 14), FIRST) || (__c4['🧨'](1, 23), SECOND).VALUE) || (__c4['🧨'](1, 40), HELLO);
```

Because if `__c4['🧨']` calls are removed, code looks like this:

```js
 const name = (FIRST || SECOND.VALUE) || HELLO;
```

Which is absolutely wrong. Here is [code](https://github.com/coderaiser/escover/blob/v1.19.0/lib/instrument/plugin-mark/index.js#L162-L170) (+ what inslide a `fix` function) of 🐊[Putout](https://github.com/coderaiser/putout) plugin that does transforms with Babel API.

coderaiser:
Here is test example:

```ts
it("sequence", async function() {
    const source = 'const name = (a || b).value || value';
    const expected = `const name = (__c4['🧨'](1, 13), ((__c4['🧨'](1, 14), a) || (__c4['🧨'](1, 19), b).value)) || (__c4['🧨'](1, 31), value)`;

    // @ts-ignore
    const plugin = await eval('import("escover/plugin")') as any;
    // @ts-ignore
    const {transform, parse, print} = await eval('import("putout")') as any;

    const ast = parse(source);

    transform(ast, source, {
        plugins: [
            ['escover', plugin],
        ],
    });

    assert.strictEqual(
      print(ast),
      expected,
    );
});

Only transformed made by 🐊`Putout`, `parse` and `print` it's recast.

PR #1068, opened 2022-03-03
Fix missing parens in reprinted output

### LINKED ISSUE #1067: ArrowFunctionExpression returning ObjectPattern prints without parens when a return type is removed
I've created a minimal repro for this behavior here: https://github.com/conartist6/recast-arrow-object-pattern-repro

The input is:
```ts
(): Type => ({ prop: true });
```
and the output is invalid, as it is missing parens:
```
() => { prop: true };
```

conartist6:
Fixes #1067

I am not sure the test is completely adequate, but I can verify that the bug is fixed in the minimal repro.

### DIFF
--- a/lib/fast-path.ts
+++ b/lib/fast-path.ts
@@ -324,9 +324,6 @@ FPp.needsParens = function (assumeExpressionContext) {
   }
 
   const parent = this.getParentNode();
-  if (!parent) {
-    return false;
-  }
 
   const name = this.getName();
 
@@ -347,12 +344,15 @@ FPp.needsParens = function (assumeExpressionContext) {
     return false;
   }
 
-  if (
-    parent.type === "ParenthesizedExpression" ||
-    (node.extra && node.extra.parenthesized)
-  ) {
+  if (parent && parent.type === "ParenthesizedExpression") {
     return false;
   }
+
+  if (node.extra && node.extra.parenthesized) {
+    return true;
+  }
+
+  if (!parent) return false;
 
   switch (node.type) {
     case "UnaryExpression":
--- a/test/parens-extra.ts
+++ b/test/parens-extra.ts
@@ -1,7 +1,9 @@
 import assert from "assert";
+// the babel parser denotes decorative parens with extra.parenthesized
 import * as babylon from "@babel/parser";
 import { parse as recastParse } from "../lib/parser";
 import { Printer } from "../lib/printer";
+import * as parser from "../parsers/babel-ts";
 import * as types from "ast-types";
 
 const printer = new Printer();
@@ -12,23 +14,20 @@ function parseExpression(expr: any) {
   return n.ExpressionStatement.check(ast) ? ast.expression : ast;
 }
 
-const parse = (expr: string) =>
-  recastParse(expr, {
-    parser: babylon,
-  });
+const parse = (expr: string) => recastParse(expr, { parser });
 
 function check(expr: string) {
   const ast = parse(expr);
 
   const reprinted = printer.print(ast).code;
-  assert.strictEqual(reprinted, expr);
+  assert.strictEqual(expr, reprinted);
 
   const expressionAst = parseExpression(expr);
   const generic = printer.printGenerically(expressionAst).code;
   types.astNodesAreEquivalent.assert(expressionAst, parseExpression(generic));
 }
 
-describe("babylon parens", function () {
+describe("parens from node.extra.parenthesized", function () {
   it("AwaitExpression", function () {
     check("async () => ({...(await obj)})");
     check("(async function* () { yield await foo })");
@@ -54,5 +53,21 @@ describe("babylon parens", function () {
     expr.object.extra.parenthesized = false;
 
     assert.strictEqual(printer.print(ast).code, "(1).foo");
+  });
+
+  it("prints top level parens for an expression ast", function () {
+    check("(() => {})()");
+    check("(function () {} ())");
+  });
+
+  describe("reprinter", function () {
+    it("preserves necessary parens", function () {
+      const ast = parse("() => ({ prop: true })");
+      const expr = ast.program.body[0].expression;
+
+      expr.body.properties = [];
+
+      assert.strictEqual(printer.print(ast).code, "() => ({})");
+    });
   });
 });
--- a/test/run.ts
+++ b/test/run.ts
@@ -5,6 +5,7 @@ import "./identity";
 import "./jsx";
 import "./lines";
 import "./mapping";
+import "./parens-extra";
 import "./parens";
 import "./parser";
 import "./patcher";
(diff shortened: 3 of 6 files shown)


conartist6:
Oh hmm, nope it's not quite right yet. I found a test that breaks things.

conartist6:
Oops, didn't mean to send that yet. The problem is that `printer.printGenerically(parseExpression('(function () {})'))` loses its outermost parens because `FPp.needsParens` returns `false` any time a node has no parent. This short-circuits both the `extra.parenthesized` logic and and the `firstInStatement` logic, either of which could indicate that the outermost parens surrounding the `FunctionExpression` are necessary. Interestingly only the typescript test happens to trigger this condition, but the underlying bug is not unique to typescript.

conartist6:
It seems like I could just tweak the logic to ensure that neither of those is dependent on `!node.parent`. Both of those expressions already seem to make perfect sense when `!node.parent`.

conartist6:
By the way this is only an issue because I've removed the `extra.parenthesized` check [here](https://github.com/benjamn/recast/blob/cb7ff9a333b68ced3a6c79be5613c7c48bbe6d8a/lib/printer.ts#L199) as it was causing paren doubling when printing code parsed with the babylon parser.

conartist6:
Though actually that line only handled the case when `extra.parenthesized` was true, which should not be a requirement for correct behavior.

conartist6:
So I think I've fixed those problems.

I also renamed `parens-babylon.ts` to `parens-extra` to make it more clear what aspect of the code is under test. I've moved some assertions about that behavior from the typescript test to this one as well, and added coverage for reprinting.

conartist6:
As it sits now this PR has one failing test that I'm not sure what to do about. The test case boils down to this:
```ts
    assert.strictEqual(
      new Printer().print(b.functionExpression(/*...*/)).code,
      "function a(b, c = 1, ...d) {}",
    );
```
which results in the following error:
```
      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ '(function a(b, c = 1, ...d) {})'
- 'function a(b, c = 1, ...d) {}'
      + expected - actual

      -(function a(b, c = 1, ...d) {})
      +function a(b, c = 1, ...d) {}
```

I definitely broke this by allowing the root node of an expression AST to need parens, and I read the result as malicious compliance. For `*Expression` nodes which have equivalent `*Statement` I do not see a downside to allowing an expression at the root of a printed AST to be interpreted as a statement as a special case.

conartist6:
Ah ok, I'm putting it back exactly the way it was. All the tests pass again now. It looks a little different because it is still necessary to check `extra.parenthesized` even if there is no parent.

That ensures that both the formerly broken case and `check("(function () {} ())");` pass.

lib/fast-path.ts
```diff
@@ -324,9 +324,6 @@ FPp.needsParens = function (assumeExpressionContext) {
   }
 
   const parent = this.getParentNode();
-  if (!parent) {
-    return false;
-  }
```
conartist6, line 329: We delay this check so that we can reprint `(function () {} ())`, which otherwise will always have its outermost parens stripped since it has no parent.

lib/fast-path.ts
```diff
@@ -347,205 +344,208 @@ FPp.needsParens = function (assumeExpressionContext) {
...
-        case "CallExpression":
-          return name === "callee" && parent.callee === node;
-
-        case "UnaryExpression":
-        case "SpreadElement":
-        case "SpreadProperty":
-          return true;
+  if (node.extra && node.extra.parenthesized) {
+    return true;
+  }
```
conartist6, line 353: The problem with short circuiting to `false` as we did previously is that a node may still really **need** parens (that is, non-decoratively). I opted to make this code responsible for decorative parens which as far as I can tell works fine, though it isn't apparent to me yet if there is some reason the code was written the way it was. The test suite seems to indicate that there is not, or at least is no longer.

test/run.ts
```diff
@@ -5,6 +5,7 @@ import "./identity";
 import "./jsx";
 import "./lines";
 import "./mapping";
+import "./parens-extra";
```
conartist6, line 8: d'oh

test/parens-extra.ts
```diff
@@ -55,4 +54,20 @@ describe("babylon parens", function () {
...
+  });
+
+  describe("reprinter", function () {
+    it("preserves necessary parens", function () {
+      const ast = parse("() => ({ prop: true })");
+      const expr = ast.program.body[0].expression;
+
+      expr.body.properties = [];
+
+      assert.strictEqual(printer.print(ast).code, "() => ({})");
```
conartist6, line 70: I haven't yet been able to verify that this test fails if the behavior regresses.

test/parens-extra.ts
```diff
@@ -12,23 +14,20 @@ function parseExpression(expr: any) {
...
-    parser: babylon,
-  });
+const parse = (expr: string) => recastParse(expr, { parser });
 
 function check(expr: string) {
   const ast = parse(expr);
 
   const reprinted = printer.print(ast).code;
-  assert.strictEqual(reprinted, expr);
+  assert.strictEqual(expr, reprinted);
```
conartist6, line 23: expected is first

conartist6 (commented):
I strongly recommend reviewing this with the "Hide whitespace" option on as a large block was reindented.

lib/fast-path.ts
```diff
@@ -347,205 +344,208 @@ FPp.needsParens = function (assumeExpressionContext) {
...
 
-        case "MemberExpression":
-          return name === "object" && parent.object === node;
+  if (parent) {
```
gnprice, line 355: Can you make this say `if (parent) { return false; }` instead?

If I'm reading this right, that would mean the remaining 200 or so lines of this function don't have to get reindented. That would make this diff a lot smaller, and easier to read and review.

It would also make the resulting code easier to read. Instead of being in suspense about what happens if `parent` is falsy -- and having to find the end of this `if` block 194 lines later to find out -- the reader would be able to see the answer immediately. Then they can move on to focusing on the interesting case, without having that question in the back of their head.

conartist6: Ah, yeah, I can. I originally did I think because I thought there was some logic at the bottom that belonged outside the `if(parent)`, but actually it turned out there wasn't.

eventualbuddha (approved):

lib/fast-path.ts
gnprice: Great :-) And I see it's now been merged!

conartist6:
@coderaiser Sorry about that. The test cases for recast are really rather inadequate it seems. I'm looking into these issues.

conartist6:
I'm looking at the extra parens on JSX return first. The JSX node itself is `node.extra.parenthesized`, and the printer always adds parens when returning a multiline JSX node as a return statement. No mystery here really.

Unfortunately it's pretty tricky to see what the correct fix may be....

conartist6:
Out of curiousity @coderaiser how do you feel about transitioning from recast to prettier if prettier were able to support transformations?

PR #1077, opened 2022-03-30
Test/ambient context

coderaiser:
When you update **@babel/parser** to latest version you will see broken tests:

image

This is related to next code:

```ts
declare namespace D3 {
  export const f: number = 2;
}
```

It has an error and should not [contain implementation](https://stackoverflow.com/questions/26946495/what-means-ambient-in-typescript):

```ts
declare namespace D3 {
  export const f: number;
}
```

Here is the same error in [TypeScript Playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEA7KBbEBnADlMCAiAzPAN4BQ88IAHpgPYwAu8Yti6TAZgFxICuKAIxAx4AXngAmANykAvqVJA).

### DIFF
--- a/test/typescript.ts
+++ b/test/typescript.ts
@@ -273,7 +273,7 @@ const nodeMajorVersion = parseInt(process.versions.node, 10);
 
     check(["export interface S {", "  i(j: string): boolean;", "}"]);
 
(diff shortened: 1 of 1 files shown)


eventualbuddha:
Fixed in v0.21.4.

PR #1090, opened 2022-04-27
Support Flow "declare class" fully: type parameters, "extends", "implements"

gnprice:
And add test cases.

The first commit does this by basically copying from the "ClassDeclaration" case.

But there's so much in common between them -- it'd be nice to share most of this logic. Otherwise, as this breakage demonstrates, it's easy for one of them to lag behind in getting a feature the other has. So in the second commit, we unify the handling of "DeclareClass" (used for Flow "declare class …") with "ClassDeclaration" and "ClassExpression", which are already unified with each other.

The "implements" property on "DeclareClass" isn't yet known to ast-types -- I'll send a PR there for that. But it's already emitted by flow-parser, so the round-trip tests already work.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1472,6 +1472,7 @@ function genericPrintNoParens(path: any, options: any, print: any) {
 
     case "ClassDeclaration":
     case "ClassExpression":
+    case "DeclareClass":
       if (n.declare) {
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Fixed in v0.21.5.

PR #1105, opened 2022-05-07
Add support for the Meriyah parser

prantlf:
[Meriyah] is a [fast] and modern implementation of ECMAScript parser.

grafik

[Meriyah]: https://meriyah.github.io/meriyah/
[fast]: https://meriyah.github.io/meriyah/performance/


.github/workflows/main.yml
```diff
@@ -2,28 +2,28 @@ name: CI
...
+    branches: [master]
 
 jobs:
   test:
     name: Test on node ${{ matrix.node_version }} and ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node_version: ['10.x', '12.x', '14.x', '15.x', '16.x']
+        node_version: ["10.x", "12.x", "14.x", "15.x", "16.x"]
```
gnprice, line 15: Is there any intentional change in this file? If not, best to leave the file unmodified, to avoid adding noise to the diff.

Same with `tsconfig.json`.

parsers/meriyah.ts
```diff
@@ -0,0 +1,36 @@
...
+//
+//   const ast = recast.parse(source, {
+//     parser: require("recast/parsers/meriyah")
+//   });
+//
+import { getOption } from "../lib/util";
+
+function convertComment(comment: any): any {
+  let { type } = comment;
+  type = type[0] === "S" ? "Line" : "Block"; // SingleLine/MultiLine
```
gnprice, line 12: Are the anticipated values of `type` literally 'SingleLine' and 'MultiLine'?

If so, clearest to directly say `type === 'SingleLine' ? …`, rather than look at `type[0]`.

gnprice (commented):
Neat that it's so much faster than Acorn or Esprima. How does the speed compare to Babel?

Small code comments below.

.github/workflows/main.yml
prantlf: Oh, I didn't notice changed in `.github/workflows/main.yml` and in `tsconfig.json`. They must've been caused by running `npm run format` before final testing and committing.

I'm reverting changes in the two files.

parsers/meriyah.ts
prantlf: Yes, will do.

prantlf:
> Neat that it's so much faster than Acorn or Esprima. How does the speed compare to Babel?

Babel is among the slowest, but [my benchmark] uses only jQuery and Marionette.CollectionView.

[my benchmark]: https://github.com/prantlf/ecmascript-parser-benchmark#parsing-speed

gnprice (approved):
Good to know, thanks. These changes look good to me! The maintainers may have other comments.

mrienstra:
Bump -- this looks like a good addition.

In case it's helpful: Meriyah is used by [astring](https://github.com/davidbonnet/astring) & [nexe](https://github.com/nexe/nexe). And indirectly by [react-scanner](https://github.com/moroshko/react-scanner), by way of `astray`.

eventualbuddha:
This isn't something I feel I can decide whether to incorporate. I'll leave it to @benjamn.

PR #1116, opened 2022-05-25
Rename prepack script to prepare, to allow installation from Git

andersk:
Like benjamn/ast-types#516.


eventualbuddha:
Released in v0.21.2.

Issue #1146, opened 2022-07-12
Curly braces of TSModuleBlock get removed when child of TSModuleBlock is removed


suchipi:
# Problem

If I use recast to remove `export const four = 4` from the following code block:
```ts
namespace Thing {
  export const four = 4;
  export const five = 5;
}
```

I expect this output:
```ts
namespace Thing {
  export const five = 5;
}
```

But instead, `recast.print` gives this output, which contains invalid syntax, as the curly braces for the namespace body have been removed:
```ts
namespace Thing export const five = 5;
```

# Info

The problem does not happen with `recast.prettyPrint`.

I encountered this bug while using the following package versions:
```json
{
  "@babel/core": "7.18.6",
  "@babel/parser": "7.18.8",
  "recast": "0.21.1"
}
```

# Repro

Here's a repro script:

```js
const recast = require("recast");
const parser = require("recast/parsers/babel-ts");

const ast = recast.parse(
  `
namespace Thing {
  export const four = 4;
  export const five = 5;
}
`,
  { parser }
);

// Remove `export const four = 4;`
ast.program.body[0].body.body = ast.program.body[0].body.body.slice(1);

const result = recast.print(ast);
console.log(result.code); // namespace Thing export const five = 5;
```

PR #1147, opened 2022-07-12
Fix disappearing TSModuleBlock curly braces

suchipi:
Fixes https://github.com/benjamn/recast/issues/1146.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2487,25 +2487,28 @@ function genericPrintNoParens(path: any, options: any, print: any) {
 
       parts.push(path.call(print, "id"));
 
-      if (n.body && n.body.type === "TSModuleDeclaration") {
+      if (n.body) {
+        parts.push(" ");
         parts.push(path.call(print, "body"));
-      } else if (n.body) {
-        const bodyLines = path.call(print, "body");
-        if (bodyLines.isEmpty()) {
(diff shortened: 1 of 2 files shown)


lib/printer.ts
```diff
@@ -2487,26 +2487,29 @@ function genericPrintNoParens(path: any, options: any, print: any) {
...
-        }
       }
 
       return concat(parts);
     }
 
-    case "TSModuleBlock":
-      return path.call(
+    case "TSModuleBlock": {
+      const naked = path.call(
```
eventualbuddha, line 2499: Seems like `contents` would be a better name?

eventualbuddha (approved):
Looks good except for a variable name.

lib/printer.ts
suchipi: Named it this because that's what is already done for `BlockStatement`: https://github.com/benjamn/recast/blob/aec4d901fafc195d4b5885ee91822e338df033a9/lib/printer.ts#L629

lily-controlzee:
Good to merge?

suchipi:
oops, made that comment from the wrong account

eventualbuddha:
Released in v0.21.2.

suchipi:
Awesome; thank you so much!

PR #1151, opened 2022-07-20
Fix typescript override keyword printer

625dennis:
When using `recast.print` function, `override` keyword is not preserved on `ClassMethod` nodes.

Reproduction link https://astexplorer.net/#/gist/a5f747da23a3539448fb63ffe9f04c43/5be0fb3874deba008cd4251e2756b8adb9289bea

Original issue on jscodeshift https://github.com/facebook/jscodeshift/issues/513


eventualbuddha:
Fixed in v0.21.4.

PR #1157, opened 2022-07-29
fix: additional commas in TsTypeLiteral

henryqdineen:
This is very pretty similar to #868 but with `TsTypeLiteral`. We were working on a codemod that involves adding to a `type` and it was resulting in syntactically incorrect code like this (notice the `;,`):

```
type FooProps = {
  bar: string;,
  children?: React.ReactNode
}
```

We were able to `patch-package` but I figured I would make an upstream fix. Thank you!

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2186,7 +2186,7 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       ]);
 
     case "TSTypeLiteral": {
-      const memberLines = fromString(",\n").join(path.map(print, "members"));
+      const memberLines = fromString("\n").join(path.map(print, "members"));
 
       if (memberLines.isEmpty()) {
         return fromString("{}", options);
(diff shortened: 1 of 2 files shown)


eventualbuddha (commented):
Thanks! It'd be very useful to have a test that fails before the changes you've made. Mind doing that?

henryqdineen:
Thanks! I just pushed a commit to add a test following a similar pattern as #1157 

Against `master` the test fails with this error:

```
  1) TypeScript
       TypeLiteral: unwanted comma:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'type A = {\n  x: boolean;,\n  y: number;\n}'
- 'type A = {\n  x: boolean;\n  y: number;\n}'
                            ^
      + expected - actual

       type A = {
      -  x: boolean;,
      +  x: boolean;
         y: number;
       }
```

eventualbuddha (approved):
Thanks! Will merge and do a release soon.

eventualbuddha:
Released in v0.21.2.

Issue #1230, opened 2022-11-24
Error: did not recognize object of type "TSInstantiationExpression"


Arooba-git:
Hi 👋 
I'm using recast with babel parser, and getting this error when trying to parse any file containing TSInstantiationExpression. Is recast missing TSInstantiationExpression support or am I missing something? 🤔

<img width="950" alt="Screen Shot 2022-11-22 at 6 15 55 AM" src="https://user-images.githubusercontent.com/56495631/203684651-c09e566f-f1bb-429e-a72b-2d8725df4623.png">

Arooba-git:
bump

iliakan:
Is there any workaround or a temporary fix for that? I'd be happy just to ignore this, just not to die.

My jscodeshift auto-refactoring script fails to work on real-life files because of that. Even though it doesn't need to refactor any TSInstantiationExpressions.

eventualbuddha:
Does it fail on _every_ file? Or only files with `TSInstantiationExpression`s? You could comment them out for now.

iliakan:
I have a large codebase, 100+ files have `TSInstantiationExpressions`s. The refactoring script fails on them.

iliakan:
Without any type of pressure, may I kindly wonder, is this going to be fixed soon?

I saw a commit has landed, but not sure how much work is ahead.

Thank you for what you do!

benjamn:
The remaining work in https://github.com/benjamn/recast/pull/1232 is worthwhile for completeness but not necessarily relevant unless you're using `ClassAccessorProperty` (a very new AST node type).

@iliakan If you want to test the current changes locally, you should be able to run this command to install `recast@0.23.0-pr1232.0`:
```sh
npm i recast@pr-1232
```
If this is useful, let me know. I might try to automate it for any PR, so it's easier for folks to test/use any proposed changes.

iliakan:
Last time it worked without errors (but I did not check the outcome). I'll retest again after I rewrite some stuff ;)

benjamn:
Alright folks, please try running `npm i recast@next` to get version 0.23.1, which includes support for `TSInstantiationExpression` and some other AST types that were lagging behind Babel.

Since I have not tagged this version as `latest` on npm yet, we can still fix bugs relatively easily, if you encounter anything strange after updating.

PR #1232, opened 2022-11-25
Add support for `TSInstantiationExpression`

eventualbuddha:
Refs #1230

See https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#instantiation-expressions. This won't work yet because `ast-types` does not yet know about `TSInstantiationExpression`. Once it does `recast` will need to be updated to include the newer release of `ast-types`, at which point it will be supported.

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -1,4 +1,3 @@
-import { Omit } from "ast-types/types";
 import { getLineTerminator } from "./util";
 
 /**
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1,14 +1,14 @@
 import assert from "assert";
+import * as types from "ast-types";
 import { printComments } from "./comments";
-import { Lines, fromString, concat } from "./lines";
+import FastPath from "./fast-path";
+import { concat, fromString, Lines } from "./lines";
 import { normalize as normalizeOptions } from "./options";
 import { getReprinter } from "./patcher";
-import * as types from "ast-types";
+import * as util from "./util";
 const namedTypes = types.namedTypes;
 const isString = types.builtInTypes.string;
 const isObject = types.builtInTypes.object;
-import FastPath from "./fast-path";
-import * as util from "./util";
 
 export interface PrintResultType {
   code: string;
@@ -1469,6 +1469,40 @@ function genericPrintNoParens(path: any, options: any, print: any) {
 
       parts.push(";");
       return concat(parts);
+
+    case "ClassAccessorProperty": {
+      if (n.static) {
+        parts.push("static ");
+      }
+
+      parts.push("accessor ");
+
+      if (n.computed) {
+        parts.push("[", path.call(print, "key"), "]");
+      } else {
+        parts.push(path.call(print, "key"));
+      }
+
+      if (n.optional) {
+        parts.push("?");
+      }
+
+      if (n.definite) {
+        parts.push("!");
+      }
+
+      if (n.typeAnnotation) {
+        parts.push(path.call(print, "typeAnnotation"));
+      }
+
+      if (n.value) {
+        parts.push(" = ", path.call(print, "value"));
+      }
+
+      parts.push(";");
+
+      return concat(parts);
+    }
 
     case "ClassDeclaration":
     case "ClassExpression":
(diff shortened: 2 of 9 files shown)


eventualbuddha:
@benjamn I can make a PR for `ast-types` for you to review, or you can give me write+publish access. Your call.

benjamn:
@eventualbuddha Thanks for offering to do the `ast-types` part too! I'm happy to give you write/publish access, but I've realized the `ast-types` repo is not in good shape for local development right now (`npm i` fails, and that's not all), so I'd like to fix that before asking anyone else to work on that project. I should have a PR up tomorrow.

benjamn:
@eventualbuddha Here's the PR I promised, though it happens to be an existing PR where @octogonz had already done most of the work! https://github.com/benjamn/ast-types/pull/327#issuecomment-1334573349

I will probably merge that PR soon, which should fix `npm install` and make it possible to add new defs again. If you still have time to add a def for `TSInstantiationExpression`, please give it a try, and let me know if you hit any snags.

iliakan:
bump =)

benjamn:
Progress! https://github.com/benjamn/ast-types/pull/886

benjamn:
Based on those [test failures](https://github.com/benjamn/recast/actions/runs/3670598533/jobs/6205246283) (newly [itemized](https://github.com/benjamn/recast/pull/1232/commits/62ffa684b7023d20daf9dad6831bab1df237abb1)!), we still need to implement pretty-printer `switch (node.type)` cases for the following the following AST types:
- [ ] [`ClassAccessorProperty`](https://github.com/babel/babel/blob/f5b52208f05157a348fdfaa0222c07a9a83fb101/packages/babel-generator/src/generators/classes.ts#L123-L166)
- [ ] [`TSTypeCastExpression`](https://github.com/babel/babel/blob/b84f4f2d69228010f86baf82d107728b860bca9d/packages/babel-generator/src/generators/typescript.js#L376-L379) (possibly no longer used?)
- [ ] [`TSSatisfiesExpression`](https://github.com/babel/babel/blob/f5b52208f05157a348fdfaa0222c07a9a83fb101/packages/babel-generator/src/generators/typescript.ts#L525-L541)

benjamn:
@eventualbuddha Wanna try implementing the pretty-printer case for [`ClassAccessorProperty`](https://github.com/babel/babel/blob/f5b52208f05157a348fdfaa0222c07a9a83fb101/packages/babel-generator/src/generators/classes.ts#L123-L166)? If not, or you run out of time, just let me know and I'll finish it up.

eventualbuddha:
Alright, I got most of the way to a pretty-printer implementation for `ClassAccessorProperty` but hit an issue. In Babel, the AST can represent a "definite" property as in this TypeScript code:

```ts
class Foo {
  accessor bar!: unknown;
}
```

However, `ast-types` is unaware of the `definite` field for this node so the type checking fails:

```
test/printer.ts:1311:15 - error TS2345: Argument of type '{ key: types.namedTypes.Identifier; definite: boolean; }' is not assignable to parameter of type '{ abstract?: boolean | undefined; accessibility?: "public" | "private" | "protected" | null | undefined; comments?: CommentKind[] | null | undefined; computed?: boolean | undefined; ... 7 more ...; value: ExpressionKind; }'.
  Object literal may only specify known properties, and 'definite' does not exist in type '{ abstract?: boolean | undefined; accessibility?: "public" | "private" | "protected" | null | undefined; comments?: CommentKind[] | null | undefined; computed?: boolean | undefined; ... 7 more ...; value: ExpressionKind; }'.

1311               definite: true,
                   ~~~~~~~~~~~~~~
```

What do you think, @benjamn?

benjamn:
@eventualbuddha Can you try updating to `ast-types@0.16.1` (just published), which includes https://github.com/benjamn/ast-types/pull/891?

eventualbuddha:
@benjamn alright, added a commit to update `ast-types` and a small refactor to consolidate printing of class member modifiers.

benjamn (approved):

benjamn:
Running `npm i recast@next` will give you version 0.23.1 (or later), which includes this change.

Thanks to @eventualbuddha for tackling this PR and for getting that final test passing.

If everything looks good here, I will tag this version as `latest` on npm in a few days.

PR #1237, opened 2022-11-29
feat: make lib browser compatible

### LINKED ISSUE #1235: Consider removing usage of Node.js builtins
I'd like to use this library in the browser and noticed it uses `assert`, and `os` from Node.js, only for `os.EOL` and some assertion functions like `ok`. While Webpack 4 polyfills Node.js builtins automatically, it's not ideal, as those polyfills often cause issues. For that reason, Webpack 5 no longer polyfills these modules. This is also the case for Vite.

I was wondering if we could remove these and use alternatives or require them but fallback to some alternatives so this package would work in the browser as well as with Node.js.

d3lm:
This PR makes recast browser compatible by using an `assert` polyfill which will only be used in the browser, because with Node.js `assert` is a built-in, as well as conditionally requires `os` only when executed with Node.js.

Closes #1235

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -1,4 +1,5 @@
 import { Omit } from "ast-types/types";
+import { getLineTerminator } from "./util";
 
 /**
  * All Recast API functions take second parameter with configuration options,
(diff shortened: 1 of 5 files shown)


d3lm:
@eventualbuddha Any chance we could get this in and published? I'd really like to use recast in the browser, and for the time being I am using my fork.

eventualbuddha (approved):

eventualbuddha:
Published in v0.22.0.

d3lm:
Thanks so much!

PR #1247, opened 2022-12-10
Restore pretty-printing TypeScript `type T = {...}` and `interface T {...}` semicolons between child members

benjamn:
PR #1157 correctly stopped the Recast pretty-printer from adding unnecessary extra commas between the fields of object-like TypeScript declarations.

However, many/most TypeScript programmers would probably expect to see _semicolons_ here instead of commas/nothing. We could make this configurable, but I'm not sure it's even controversial, so I'll wait for someone to ask.

This PR enables the Recast pretty-printer (which is only invoked when the AST changes enough to require reprinting) to print those semicolons properly.

The subtlety is to avoid printing _extra_ semicolons, but we already have precedent for using `lastNonSpaceCharacter` to determine what's already at the end of printed `Lines` output.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2428,7 +2428,14 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       ]);
 
     case "TSInterfaceBody": {
-      const lines = fromString("\n").join(path.map(print, "body"));
+      const lines = fromString("\n").join(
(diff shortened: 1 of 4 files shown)


eventualbuddha (approved):
I had wondered why some of the generated code I was seeing in `ant-types` no longer had semicolons. Maybe this was why?

Issue #1248, opened 2022-12-12
Extra parens inserted around JSX after attributes are added


michaldudak:
(This was originally posted in jscodeshift's repo: https://github.com/facebook/jscodeshift/issues/534, but I was redirected to recast)

**Context**
We're using jscodeshift (and, thus, recast) to provide codemods that help our users migrate to new MUI libraries versions. After updating jscodeshift to 0.14.0 (which updates recast to 0.21.0), our tests started to fail. 
It seems that when a codemod adds a JSX attribute, the whole JSX block in a return statement gets wrapped in extra parens (even if it had parens before).

I managed to create a minimal repro:

**index.js**
```js
const fs = require("fs");
const recast = require("recast");
const { builders: b, visit } = require("ast-types");

const source = fs.readFileSync("source.js", "utf8");

const ast = recast.parse(source, {
  parser: require("recast/parsers/babel"),
});

visit(ast, {
  visitJSXElement: function (path) {
    let attributes = path.node.openingElement.attributes;
    attributes.push(
      b.jsxAttribute(b.jsxIdentifier("data-foo"), b.literal("bar"))
    );

    this.traverse(path);
  },
});

console.log(recast.print(ast).code);
```

**source.js**
```jsx
import * as React from "react";

export default function WrappedLink(props) {
  return (
    <div>
      <a {...props} />
    </div>
  );
}
```

**expected output**
```jsx
import * as React from "react";

export default function WrappedLink(props) {
  return (
    <div data-foo="bar">
      <a {...props} data-foo="bar" />
    </div>
  );
}
```

**actual output**
```jsx
import * as React from "react";

export default function WrappedLink(props) {
  return (
    (<div data-foo="bar">
      <a {...props} data-foo="bar" />
    </div>)
  );
}
```

JosephBrooksbank:
Looks like a fix has been up and ready for review for 2 weeks. Michał did you find a workaround in the meantime?

eventualbuddha:
> Looks like a fix has been up and ready for review for 2 weeks. Michał did you find a workaround in the meantime?

#1257 isn't really a fix, is it? A proper fix would be to not add the double parens in the first place, right?

michaldudak:
> Michał did you find a workaround in the meantime?

We haven't upgraded to the broken version.

JosephBrooksbank:
> We haven't upgraded to the broken version.

Which version of jscodeshift and recast are you using?

michaldudak:
[...]

PR #1257, opened 2022-12-16
[Fix] Add avoid parens config for Printer

### LINKED ISSUE #1191: `recast.print` modifies original source adding brackets around some statements
Ref: https://github.com/facebook/jscodeshift/issues/176

For e.g. Let's consider the following source code as input passed to `recast.print()`;

```js
const code = [
  "async function add(a, b) {",
  "  const z = foo || await getFoo();",
  "  return z;",
  "}"
].join("\n");

const ast = recast.parse(code);
recast.print(ast);
```

## Behaviour:

Prints a modified code, with `await getFoo()` surrounded by parentheses (normal brackets).
```js
// PrintResult {
//   code: 'async function add(a, b) {\n' +
//     '  const z = foo || (await getFoo());\n' +
//     '  return z;\n' +
//     '}'
// }
```

## Expected Behaviour:

Should not contain parenthesis if they were not present in the original code.
```js
// PrintResult {
//   code: 'async function add(a, b) {\n' +
//     '  const z = foo || await getFoo();\n' +
//     '  return z;\n' +
//     '}'
// }
```

If there are any restrictions in supporting this, can it be possible to put this _"adding auto parentheses"_ feature behind a recast `options` toggle?

Also if you need any support, I can work on this issue with some guidance 👍🏽 
Thanks for all your creating this awesome library 🙇🏽 💯

phoenisx:
Fixes #1191, and possibly #1248

### DIFF
--- a/lib/options.ts
+++ b/lib/options.ts
@@ -157,6 +157,12 @@ export interface Options extends DeprecatedOptions {
    * @default true
    */
   tokens?: boolean;
+
+  /**
+   * Whether to add parens around statements when possible while printing the AST.
+   * @default false
+   */
+  avoidParens?: boolean;
 }
 
 interface DeprecatedOptions {
@@ -184,6 +190,7 @@ const defaults: Options = {
   arrowParensAlways: false,
   flowObjectCommas: true,
   tokens: true,
+  avoidParens: false,
 };
 const hasOwn = defaults.hasOwnProperty;
 
@@ -219,5 +226,6 @@ export function normalize(opts?: Options): NormalizedOptions {
     arrowParensAlways: get("arrowParensAlways"),
     flowObjectCommas: get("flowObjectCommas"),
     tokens: !!get("tokens"),
+    avoidParens: !!get("avoidParens"),
   };
 }
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -202,7 +202,7 @@ function genericPrint(path: any, config: any, options: any, printPath: any) {
   if (decoratorsLines.isEmpty()) {
     // Nodes with decorators can't have parentheses, so we can avoid
     // computing path.needsParens() except in this case.
-    if (!options.avoidRootParens) {
+    if (!options.avoidRootParens && !config.avoidParens) {
       shouldAddParens = path.needsParens();
     }
   } else {
--- a/test/printer.ts
+++ b/test/printer.ts
@@ -2565,4 +2565,32 @@ describe("printer", function () {
       ),
     );
   });
+
+  const withoutParens = [
+    "async function add(a, b) {",
+    "  const z = foo || await getFoo();",
+    "  return z;",
+    "}"
+  ].join(eol);
+
+  const withParens = [
+    "async function add(a, b) {",
+    "  const z = foo || (await getFoo());",
+    "  return z;",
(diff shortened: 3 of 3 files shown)


GianlucaGuarini:
This Patch could help us to upgrade recast in our @riotjs/compiler. 
@eventualbuddha have had any chance to check this? How can we help you further?

techLeadGuy:
This will also fix an issue in jscodeshift, namely [https://github.com/facebook/jscodeshift/issues/566](https://github.com/facebook/jscodeshift/issues/566). How can we push this PR ?

eventualbuddha:
A quick read of this makes me concerned that enabling this flag will, in some cases, cause printing of invalid or incorrect code. Unless `needsParens` doesn't actually mean _need_ in any case?

techLeadGuy:
@eventualbuddha I think you are right, for example when you go from 1 argument in an arrow fct to no/multiple arguments, in which case you want the final code to have auto-added parens even in the case this flag would be set to true. 

I guess a better solution would be to just fix the FPp.needsParens method for the cases where parens can be bypassed (if the input code did not have them in the first place). For the await case inside conditional expression, I think it's because of [this line](https://github.com/benjamn/recast/blob/2bb3a4ffddf038138fefa8c49158ffcfe14515ea/lib/fast-path.ts#L460). 

Unless I'm missing something, I'll be as easy as removing L451 from the printer. If you agree with this approach, I'm happy to create a PR for it.

phoenisx:
That's a very good find.
Just curious, apart from arrow-function params are there any other syntax where adding parens is required after a code change?

What I am trying to understand here is why adding parens is a generic behaviour rather than being specific to such special cases.

techLeadGuy:
@phoenisx Just by looking at recast's printer logic ([this line](https://github.com/benjamn/recast/blob/2bb3a4ffddf038138fefa8c49158ffcfe14515ea/lib/printer.ts#L1813)), there's the case where you add type definitions for params that would also need to enforce parens.
```typescript
const myFunc = name => console.log(name);
// becoming
const myFunc = (name: string) => console.log(name);
```
My feeling is that, since for the most part, the parens logic is handled appropriately ( keeping existing parens in place, only adding parens when the transformation requires for syntax validity and not adding extra parens when not required ), our specific issues are more exceptions than rules.

michaelfaith:
> My feeling is that, since for the most part, the parens logic is handled appropriately ( keeping existing parens in place, only adding parens when the transformation requires for syntax validity and not adding extra parens when not required ), our specific issues are more exceptions than rules.

Or the logic to detect when parens are required isn't entirely sound.  I.e. getting triggered by jsx elements: https://github.com/benjamn/recast/issues/1248

techLeadGuy:
@michaelfaith Unfortunately not. Tried yesterday to fiddle around with recast locally, hoping to find an easy fix for all these parens related issues but there are a lot more corner cases than I originally expected and a lot of tests for them, which is great, but time consuming to properly change. Someone with a better understanding of all these cases could perhaps be a lot more efficient than myself, but as of right now, don't know when I will have the time to try to fix these again.

topwood:
need this config too，pls fix this soon

KernelDeimos:
August of last year... *oof*

PR #1314, opened 2023-06-06
package output as `esm` + `cjs`

ivanhofer:
I need this package in an `esm` context so I took the time to convert it.
There weren't that many things to change, so the process was straigt forward.

 - The `build` script now outputs `cjs` and `esm` code.
 	In order to make the code run in an `esm` context without setting `"type": "module"` inside `package.json` all generated `esm` files need to end with the `.mjs` file extension. Therefor the `fix-esm-output.js` script is needed.
 - All imports were altered using the "node16" module resolution.
 - All `require` statements are rewritten using `import` statements.
 	This did remove the "lazy-import" behavior as dynamic `import` statements must use `async-await`. But the only place where this affects regular runtime code is where `esprima` get's used. But since it is a regular dependency of `recast` nothing should fail.
 - Tests were left unchanged and still run sucessfully for the `cjs` output.

I hope this could be merged into the main project so I don't have to maintain a copy of it.
Let me know if there are still some things missing.


ivanhofer (commented):
Some `TypeScript` related issues popped up when converting from `require` (which has no type support) to `import`. The issues are already present in the current `master` branch. I marked them with `// @ts-expect-error`.

benmccann:
Unfortunately Ivan passed away while competing during a race: https://github.com/ivanhofer/typesafe-i18n/issues/739

I still get a bit sad everytime I come across a PR from him. You will be missed Ivan.

I thought I'd make a note of it here as it's likely you may be unaware and this PR has some merge conflicts that will never be addressed by the original author.

eventualbuddha:
@benmccann I’m so sorry to hear that. Thank you for notifying us. I’ll close this PR, but I hope someone else may base their future work on top of it.

package.json
```diff
@@ -21,14 +21,31 @@
   "license": "MIT",
   "main": "main.js",
   "types": "main.d.ts",
+  "module": "esm/main.mjs",
```
benmccann, line 24: this isn't necessary and will be ignored

more importantly, I just learned that tsc doesn't enforce valid output when you set `node16` unless you also have `"type": "module"`

PR #1338, opened 2023-07-05
fix: print ExportSpecifier.exportKind

ashsearle:
This fixes a small oversight: `ExportSpecifier` supports `exportKind` in a similar fashion to `ImportSpecifier` supporting `importKind`, but the print logic was missing.

Without this change it's impossible to generate an ExportNamedDeclaration containing a mix of type and value exports.

See this [broken jscodeshift example in AST Explorer](https://astexplorer.net/#/gist/38d63e7f66c1d7a14014716483fca861/b50b4c4d23601c2f07bd634a3f9c0bd832b5d496)

This is in the spirit of changes described in the original [type-only PR for TypeScript](https://github.com/microsoft/TypeScript/pull/45998), as well as changes for [SWC in 1.2.92](https://github.com/swc-project/swc/pull/2309)

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -483,6 +483,9 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       return concat(parts);
 
     case "ExportSpecifier":
+      if (n.exportKind && n.exportKind !== "value") {
+        parts.push(n.exportKind + " ");
+      }
       if (n.local) {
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):
Thanks!

eventualbuddha:
Published in v0.23.3.

Issue #1357, opened 2023-08-05
missing parentheses in exponentiation with unary left side


j4k0xb:
reproduction: https://stackblitz.com/edit/stackblitz-starters-9gkcmv?file=index.js&view=editor
```js
const { parse, print, types } = require('recast');
const b = types.builders;

const { code } = print(
  b.binaryExpression(
    '**',
    b.unaryExpression('-', b.numericLiteral(1)),
    b.numericLiteral(2)
  )
);

console.log(code);
console.log(eval(code));
```
it prints `-1 ** 2` which is invalid syntax:
> Uncaught SyntaxError: Unary operator used immediately before exponentiation expression. Parenthesis must be used to disambiguate operator precedence

expected: `(-1) ** 2`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_precedence#table 2.

related but not fixed by #363

eventualbuddha:
Thanks for reporting this, @j4k0xb! Mind trying #1361 to see if that fixes it for you?

j4k0xb:
yes that fixed it, thanks!

j4k0xb:
just found similar issue for parsing: https://stackblitz.com/edit/stackblitz-starters-guryj8?file=index.js&view=editor
```js
acorn.parse('(-1) ** 2'); // works
recast.parse('(-1) ** 2', { parser: acorn }); // errors
```

eventualbuddha:
I'm not sure why it's hardcoded to `ecmaVersion: 6`. I'm guessing it was added when that was the latest standard.

PR #1361, opened 2023-08-10
fix(print): wrap unary exprs inside `**` in parens

eventualbuddha:
Closes https://github.com/benjamn/recast/issues/1357


eventualbuddha:
Fixed in v0.23.4.

PR #1383, opened 2024-01-03
Update printer.ts to support `kind = 'get'/'set'` on `TSMethodSignature`

dandean:
`TSMethodSignature` has a `kind` property which can be set to `method` (the default), `get`, and `set`. When `get` and `set`, the method is a getter or a setter, and as such, should be prefixed with either `get ` or `set `.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -932,7 +932,7 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       return fromString(getPossibleRaw(n) || n.value + "m", options);
 
     case "StringLiteral":
-        return fromString(nodeStr(n.value, options));
+      return fromString(nodeStr(n.value, options));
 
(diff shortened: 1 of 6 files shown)


coderaiser:
Could you please add test for this? Or provide code example

coderaiser:
Added tests https://github.com/benjamn/recast/pull/1401 and landed in [`@putout/recast@1.16.0`](https://github.com/putoutjs/recast/releases/tag/v1.16.0)

dandean:
@coderaiser Apologies for the delayed response, and thank you for adding the tests!

coderaiser:
Some checks are failing in my PR, so if you fix this in your PR, I think it can make things go a bit faster :)...

eventualbuddha (approved):

eventualbuddha:
Published in v0.23.8.

PR #1387, opened 2024-03-01
feat: use invariant for assertions

43081j:
This replaces `assert` with `tiny-invariant` in the production code (`lib/`) while continuing to use node's built-in `assert` for tests.

The result is a much leaner dependency tree (as shown in the lockfile diff).

part of on ongoing effort to reduce the footprint of various packages in the ecosystem FYI

also, i ran `npm run format` which has reformatted some missed formatting inconsistencies in unrelated code i haven't otherwise touched

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   test:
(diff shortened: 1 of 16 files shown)


eventualbuddha:
Thanks @43081j! I ended up creating #1388 to make the same change but without the formatting changes. If you're still interested in making those changes you can rebase this PR or create a new one. The `tiny-invariant` change is published in v0.23.5.

43081j:
Awesome, thanks for taking a look so soon!

I'll take a look tomorrow at repurposing this to be a formatting pr since it would still be useful for future contributors (to keep it consistent)

PR #1388, opened 2024-03-01
refactor: replace `assert` with `tiny-invariant`

eventualbuddha:
This is the same as #1387, but without the unrelated formatting changes. When bundled using `esbuild`, this brings the file size down from 852KB to 712KB, a savings of 140KB (16%).


eventualbuddha:
Published in v0.23.5.

PR #1396, opened 2024-04-01
printer: Support \n\n while printing JSXElement children

JHilker:
The printer seems to treat any number of newlines within a `JSXText` as a single newline. This causes any transform touching these elements to remove newlines that tooling such as Prettier would otherwise not remove.

Though I'm a little surprised to not see [the same formatting difference in ASTExplorer](https://astexplorer.net/#/gist/bb270b6009a00f697349916d90e15b2c/092182e6c8fc9c93f63a3811db9e5cc02bf10584), though it's running on 0.21.1 instead of the latest version (0.23.6), so maybe this was a regression at some point? I can reproduce it both within the test suite, and transforms using 0.23.6.

Edit: Oh, this was because by default the transform itself was using `parsers.esprima`. It's reproducible if you swap to `parsers.babel`, [example](https://astexplorer.net/#/gist/bb270b6009a00f697349916d90e15b2c/bb9a508b0f702b0e51ffca94e67295370e3e66d8).

Current behavior:

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tr>
<td>

```jsx
const SomeComponent = () => {
  return (
    <div>
      <Foo />

      <Bar />
    </div>
  );
}
```

</td>
<td>

```jsx
const SomeComponent = () => {
  return (
    <div>
      <FooMutated />
      <Bar />
    </div>
  );
}
```
</td>
</tr>
</table>

This PR updates the printer for `JSXElement` / `JSXFragment` to return two newlines instead of a single newline for any string children which include two adjacent newlines. 

Proposed behavior:

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tr>
<td>

```jsx
const SomeComponent = () => {
  return (
    <div>
      <Foo />

      <Bar />
    </div>
  );
}
```

</td>
<td>

```jsx
const SomeComponent = () => {
  return (
    <div>
      <FooMutated />

      <Bar />
    </div>
  );
}
```
</td>
</tr>
</table>

---

Thanks in advance for any review. This seems like an innocent enough change, but I'd entirely understand if it's more complicated than I'm imagining. Open to feedback and happy to iterate on this PR!

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1315,6 +1315,8 @@ function genericPrintNoParens(path: any, options: any, print: any) {
           ) {
             if (/\S/.test(child.value)) {
               return child.value.replace(/^\s+|\s+$/g, "");
+            } else if (/\n\n/.test(child.value)) {
+              return "\n\n";
             } else if (/\n/.test(child.value)) {
               return "\n";
             }
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1339,7 +1339,7 @@ function genericPrintNoParens(path: any, options: any, print: any) {
           ) {
             if (/\S/.test(child.value)) {
               return child.value.replace(/^\s+/g, "");
-            } else if (/\n\n/.test(child.value)) {
+            } else if (/\n\s*\n/.test(child.value)) {
               return "\n\n";
             } else if (/\n/.test(child.value)) {
               return "\n";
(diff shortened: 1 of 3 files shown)


lib/printer.ts
```diff
@@ -1315,6 +1315,8 @@ function genericPrintNoParens(path: any, options: any, print: any) {
           ) {
             if (/\S/.test(child.value)) {
               return child.value.replace(/^\s+|\s+$/g, "");
+            } else if (/\n\n/.test(child.value)) {
```
tim-dev, line 1318: Fairly pedantic, but this won't work if there is any whitespace on the second empty line. You might consider making the regex something like `/\n[ \t]*\n/`

eventualbuddha:
Thanks! Published in v0.23.18.

PR #1397, opened 2024-04-08
printer: Support type parameters on JSX components in TSX

PartMan7:
Supports actually parsing cases like `<Foo<Bar> />` instead of crashing.

Was unsure how to test this, since the existing code inherits from babel-parser's fixtures, and that doesn't currently have any tests of the sort (though it is a supported format).

Please let me know if there's any changes / modifications needed to the PR!

[Demonstration on Babel REPL](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&corejs=3.21&spec=false&loose=false&code_lz=GYVwdgxgLglg9mABAERgNwDwBVEFMAeUuYAJgM6IDkAhgEYSUB8iAFAN6IAOATnJ4gF8AXIg48-InAICUogFCJE3XFBDckGEukQQANtTJkActQC2uALxs9B42dwDEAekYBuOQLlyMqTGSjcMGAA5sw2hibmFgBE0c5uckA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=react%2Ctypescript%2Cflow&prettier=false&targets=&version=7.23.2&externalPlugins=&assumptions=%7B%7D)

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1331,6 +1331,8 @@ function genericPrintNoParens(path: any, options: any, print: any) {
 
     case "JSXOpeningElement": {
       parts.push("<", path.call(print, "name"));
+      const typeDefPart = path.call(print, "typeParameters");
+      if (typeDefPart.length) parts.push(typeDefPart);
       const attrParts: any[] = [];
 
       path.each(function (attrPath: any) {
--- a/test/run.ts
+++ b/test/run.ts
@@ -14,4 +14,5 @@ import "./printer";
 import "./syntax";
 import "./flow";
 import "./typescript";
+import "./tsx";
 import "./visit";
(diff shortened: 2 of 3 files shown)


eventualbuddha:
I'd add a test to `test/jsx.ts`.

PartMan7:
`test/jsx.ts` is currently only running with babel and esprima, which (please correct me if I'm wrong) don't support TypeScript syntax (eg: tests crash even with `<b>{'a' as string}</b>`, which is valid and supported TSX). From what I could tell, the TSX tests run for recast are limited to those directly taken from babel-parser's fixtures - could you please point me in the right direction?

eventualbuddha:
I guess I'd probably create a `test/tsx.ts` file that parses with the `typescript` option and put the tests in there.

PartMan7:
Hope this works!

eventualbuddha:
Thanks! Published in v0.23.7.

PR #1398, opened 2024-04-08
printer: Don't break on inline comments in JSX

PartMan7:
Allows rewriting JSX with inline comments without potentially generating invalid code.

Source Code:
```tsx
<Foo // comment
  prop1
  prop2
/>
```

Old Output:

```tsx
<Foo // comment prop1 prop2 />
```

New Output:

```tsx
<Foo // comment
  prop1
  prop2
/>
```

### DIFF
--- a/test/jsx.ts
+++ b/test/jsx.ts
@@ -22,6 +22,7 @@ for (const { title, parser } of [
     it("should parse and print attribute comments", function () {
       check("<b /* comment */ />");
       check("<b /* multi\nline\ncomment */ />");
+      check("<b // test comment\n/>")
     });
 
     it("should parse and print child comments", function () {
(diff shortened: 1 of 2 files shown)


test/jsx.ts
```diff
@@ -22,6 +22,7 @@ for (const { title, parser } of [
     it("should parse and print attribute comments", function () {
       check("<b /* comment */ />");
       check("<b /* multi\nline\ncomment */ />");
+      check("<b // test comment\n/>")
```
eventualbuddha, line 25: This test passes even without the changes in `lib/printer.ts`, which makes sense because this code does not contain a `TSTypeParameterInstantiation` or `TSTypeParameterDeclaration`.

image

PR #1399, opened 2024-05-20
perf: avoid excessive linear scanning for large functions

JoostK:
This commit restores the token indices for a node after they have been refined
to span the node, instead of before the refinement. This avoids the potential for
excessive searches for token indices, which is implemented using a linear scan.

It was attempted to convert the linear scans into binary searches (which an
accompanying comment on `findTokenRange` indicates shouldn't be needed) and that
also avoids the problem, but using the refined positions as implemented in this
commit is even more effective, to the point where using a binary search is less
efficient in my testing.

This change can achieve a substantial improvement, where an ~2.6MB file used to be
parsed and cloned in ~120s, now reduced to ~1.5s, almost two orders of magnitude
faster.

The perf test that tests the `backbone.js` file shows that performance improves
from ~60ms to ~53ms, only a slight improvement.

---

The offending file I used for testing was [ElkJS](https://github.com/kieler/elkjs) bundled into a single file: [elkjs.js.zip](https://github.com/benjamn/recast/files/15379902/elkjs.js.zip). It used to take ~120s on a MacBook Pro M1 Max to clone the parsed tree, which has now been reduced to 1.2s for a 100x improvement.

### DIFF
--- a/lib/parser.ts
+++ b/lib/parser.ts
@@ -154,15 +154,25 @@ const TreeCopier = function TreeCopier(
 
 const TCp: TreeCopierType = TreeCopier.prototype;
 
+const originalProperty: PropertyDescriptor = {
+  configurable: false,
+  enumerable: false,
+  writable: true,
+};
+
 TCp.copy = function (node) {
+  if (typeof node !== "object") {
+    return node;
+  }
+
   if (this.seen.has(node)) {
     return this.seen.get(node);
   }
 
   if (isArray.check(node)) {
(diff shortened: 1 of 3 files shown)


eventualbuddha:
Thanks, and sorry for leaving this open for so long. I reproduced the perf improvement, but didn't try a file quite as large as you did. On a 770KB bundle I did see about a 15x speedup, so your claim makes sense.

eventualbuddha:
This is published in v0.23.17.

PR #1404, opened 2024-06-12
fix: TSTypeAnnotation printer when parent is a TSFunction type

henryqdineen:
This is very similar to https://github.com/benjamn/recast/pull/815 but with a test that properly recreates the issue. I ran into the same issue today and needed to use `patch-package` to fix `recast` in my project. 

It appears that `TSTypeAnnotation` printer code does not run when you `print()` a AST with `TSTypeAnnotation` nodes unless you replace/modify a `TSTypeAnnotation` and change the `typeAnnotation` property to be a node of a different type. In my test I switch the `typeAnnotation` from a `TSTypeReference` to a `TSQualifiedName`. 

Thanks!

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2307,8 +2307,16 @@ function genericPrintNoParens(path: any, options: any, print: any) {
     case "TSNonNullExpression":
       return concat([path.call(print, "expression"), "!"]);
 
-    case "TSTypeAnnotation":
-      return concat([": ", path.call(print, "typeAnnotation")]);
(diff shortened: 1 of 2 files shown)


henryqdineen:
Hey @benjamn! It's been a while. Sorry for the ping. Is there any change you are accepting new maintainers? We use `ast-types` and `recast` pretty heavily and are finding ourselves needing to patch it for fixes like this. Ideally we would want to contribute back to open source.

eventualbuddha:
Thanks! Published in v0.23.19.

PR #1416, opened 2024-12-11
fix: TSArrayType printer when element is a TSUnionType

henryqdineen:
When a `TSUnionType` has an `elementType` of `TSUnionType` the brackets would only end up on the last member of the union (e.g `number | string[]` instead of `(number | string)[]`). Parsers seem to wrap the `TSUnionType` in a `TSParenthesizedType` which would get around this issue but in codemods and other scenarios where nodes are built by hand it could be easy to make this mistake.

It's possible that this is just a bug in `ast-types` and a `TSUnionType` should not be a valid `elementType`.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -2122,6 +2122,9 @@ function genericPrintNoParens(path: any, options: any, print: any) {
       return fromString("never", options);
 
     case "TSArrayType":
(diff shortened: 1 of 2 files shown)


eventualbuddha:
Thanks! Released in v0.23.16.

PR #1430, opened 2026-06-29
fix: add parentheses around operands of optional member/call expressions

### LINKED ISSUE #1410: Missing parenthesis for AwaitExpression inside an optional chain
I was running into an issue with `jscodeshift`, which seems to be similar to #589. I was seeing this in astexplorer.net, which is using some older versions, but I was observing the same behaviour when running the transform with `npx jscodeshift`, which AFAICT should be using the latest `recast` version.

Here is a reproduction: https://astexplorer.net/#/gist/19cd83009b20b00367a24ded3c031eb2/d25151fcdcea6a1d24788c3606799b579eaa6f14.

When wrapping a function call with an AwaitExpression, the serialized output differs when this is part of an MemberExpression vs. an OptionalMemberExpression.

```js
// before:
asyncFn().filter(x => x);
// after:
(await asyncFn()).filter(x => x); // correct! ✅

// before:
asyncFn()?.filter(x => x);
// after:
await asyncFn()?.filter(x => x); // wrong! ❌

```

mahirhir:
`needsParens` decides whether an expression needs parentheses based on its parent. For `MemberExpression` and `CallExpression` parents it correctly parenthesizes a lower-precedence operand, but it never did the same for their optional-chaining counterparts, `OptionalMemberExpression` and `OptionalCallExpression`.

As a result, a transform that builds (or, as in #1410, mutates) an optional chain whose base is a lower-precedence expression drops the required parentheses and changes the meaning of the code:

```js
// start: asyncFn()?.filter(x => x), then wrap asyncFn() in `await`

// before
await asyncFn()?.filter(x => x)    // parses as await (asyncFn()?.filter(...))
// after
(await asyncFn())?.filter(x => x)
```

The same was true for every low-precedence operand of an optional member or call, for example `(a = b)?.c`, `(a + b)?.c`, `(a ? b : c)?.d` and `(-a)?.b`, all of which printed without the parentheses.

The fix mirrors the existing non-optional cases. In the three `needsParens` branches that already handle `MemberExpression` / `CallExpression` parents, the optional twins are now treated the same way. The `UnaryExpression` / `SpreadElement` branch uses `n.MemberExpression.check(parent)`, which already matches `OptionalMemberExpression` (the `ArrowFunctionExpression` branch relies on this too).

Continuing chains are unaffected, since their links are plain identifiers or optional members rather than low-precedence expressions: `a?.b?.c`, `a?.b.c` and `a?.b()` still print without extra parentheses.

I added a `test/parens.ts` case covering the await/assignment/binary operands, the #1410 reproduction, and the continuing-chain guard. The existing `parens` tests still pass.

Fixes #1410

### DIFF
--- a/lib/fast-path.ts
+++ b/lib/fast-path.ts
@@ -370,8 +370,10 @@ FPp.needsParens = function (assumeExpressionContext) {
     case "UnaryExpression":
     case "SpreadElement":
     case "SpreadProperty":
+      // `n.MemberExpression.check` also matches `OptionalMemberExpression`, so
+      // e.g. `(-a)?.b` is parenthesized just like `(-a).b`.
       return (
-        parent.type === "MemberExpression" &&
+        n.MemberExpression.check(parent) &&
         name === "object" &&
         parent.object === node
       );
@@ -379,7 +381,11 @@ FPp.needsParens = function (assumeExpressionContext) {
     case "BinaryExpression":
     case "LogicalExpression":
       switch (parent.type) {
+        // `OptionalCallExpression` / `OptionalMemberExpression` are the
+        // optional-chaining counterparts of `CallExpression` /
+        // `MemberExpression` and parenthesize their operands the same way.
         case "CallExpression":
+        case "OptionalCallExpression":
           return name === "callee" && parent.callee === node;
 
         case "UnaryExpression":
@@ -388,6 +394,7 @@ FPp.needsParens = function (assumeExpressionContext) {
           return true;
 
         case "MemberExpression":
+        case "OptionalMemberExpression":
           return name === "object" && parent.object === node;
 
         case "BinaryExpression":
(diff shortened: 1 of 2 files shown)


eventualbuddha (changes requested):
Thanks for the PR! Can you ensure that the tests fail without your new `case` additions? I tried removing a couple and the tests still pass.

mahirhir:
You're right, thanks for checking. Two of the new case labels weren't exercised by any assertion:

- The `UnaryExpression`/`SpreadElement`/`SpreadProperty` case (the `n.MemberExpression.check(parent)` change) — nothing tested a unary/spread node as the object of an optional chain, e.g. `(-a)?.b`.
- The `OptionalCallExpression` case under `BinaryExpression`/`LogicalExpression` — nothing tested a binary/logical node as the callee of an optional call, e.g. `(a + b)?.()`.

I confirmed by reverting just those two pieces and rerunning `test/parens.ts`: all 30 tests still passed, same as what you saw. Added assertions for both (`(-a)?.b` and `(a + b)?.()`), reverted the same two pieces again to confirm the new assertions actually fail without them, then restored the fix. All tests pass with everything back in place.

eventualbuddha:
Published in v0.23.20.

PR #1431, opened 2026-07-08
fix: parenthesize lower-precedence tagged template tags

spokodev:
The tag of a `TaggedTemplateExpression` occupies a MemberExpression/CallExpression position, so a lower-precedence tag must be parenthesized. `needsParens` had no case for it, so printing such a tag dropped the parentheses and changed meaning:

```
input:  (a || b)`x`
output: a || b`x`      // parses as a || (b`x`)
```

`(a++)`x`` even becomes invalid syntax. This adds a tag-position check returning true for the low-precedence operand types (Unary, Update, Binary, Logical, Conditional, Assignment, Await, Yield, Arrow).

### DIFF
--- a/lib/fast-path.ts
+++ b/lib/fast-path.ts
@@ -366,6 +366,30 @@ FPp.needsParens = function (assumeExpressionContext) {
     return true;
   }
 
+  // The tag of a tagged template literal occupies a MemberExpression /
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):

eventualbuddha:
Published in v0.23.14.

Issue #1362, opened 2023-08-11
CallExpression with UpdateExpression not being wrapped with parentheses


MeguminSama:
Take the input `var t = (Ae++).toString()`

With this (via acorn) we get an AST of
```json
{
      "type": "VariableDeclaration",
      "start": 2,
      "end": 25,
      "declarations": [
        {
          "type": "VariableDeclarator",
          "start": 6,
          "end": 25,
          "id": {
            "type": "Identifier",
            "start": 6,
            "end": 7,
            "name": "t"
          },
          "init": {
            "type": "CallExpression",
            "start": 8,
            "end": 25,
            "callee": {
              "type": "MemberExpression",
              "start": 8,
              "end": 23,
              "object": {
                "type": "UpdateExpression",
                "start": 9,
                "end": 13,
                "operator": "++",
                "prefix": false,
                "argument": {
                  "type": "Identifier",
                  "start": 9,
                  "end": 11,
                  "name": "Ae"
                }
              },
              "property": {
                "type": "Identifier",
                "start": 15,
                "end": 23,
                "name": "toString"
              },
              "computed": false,
              "optional": false
            },
            "arguments": [],
            "optional": false
          }
        }
      ],
      "kind": "var"
    }
```

If we run this through recast.print, we get `var t = Ae++.toString()` which is invalid syntax.

It seems to be missing the parentheses.

Tested on recast 0.21.2 and 0.23.4

MeguminSama:
Turns out acorn has a `preserveParens` to add a non-standard params value. This fixes `recast.print`.

PR #1432, opened 2026-07-08
fix: parenthesize UpdateExpression in member and callee positions

spokodev:
`needsParens` had no case for `UpdateExpression`, so an increment/decrement used as a MemberExpression object or a Call/New callee printed without parentheses:

```
(a++).b  ->  a++.b     // invalid syntax
(a++)()  ->  a++()     // invalid syntax
(++a).b  ->  ++a.b     // parses as ++(a.b), a semantics change
```

This adds an `UpdateExpression` case mirroring the existing low-precedence-operand handling. (Recast-side of issue #1362, which was closed via an acorn `preserveParens` workaround rather than a fix.)

### DIFF
--- a/lib/fast-path.ts
+++ b/lib/fast-path.ts
@@ -376,6 +376,23 @@ FPp.needsParens = function (assumeExpressionContext) {
         parent.object === node
       );
 
+    case "UpdateExpression":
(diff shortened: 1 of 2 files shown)


eventualbuddha (approved):

eventualbuddha:
Published in v0.23.14.

PR #1433, opened 2026-07-20
fix: correct source map columns for reused code that gets dedented

### LINKED ISSUE #1402: Invariant Failed - Sourcemaps
Hi there,
I am getting a recast error into the Riot.js compiler similar to this [old one](https://github.com/benjamn/recast/issues/831)

I could reproduce it with this example:
https://runkit.com/gianlucaguarini/recast-bug-2

It seems that the sourcemaps generation fails in some cases when the code gets wrapped into round brackets.

Any help is really appreciated, thank you!

Kal-Aster:
Fixes #1402.

#### DISCLAIMER

I don't know the opinion about AI usage in this project, however I don't know anything about how this works and this bug is sitting in here since a long time, so I used AI to produce a fix. I ensured that the code was matching the rest of the code in both style and structure, but other than that I just prompted the AI.

Follows an explanation of AI about the bug and its fix:

`Lines#getIndentAt` never reports a negative indentation, so a line already at column zero
stays put while the lines around it are dedented. `Mapping#indent` shifted every mapping by
the full requested amount anyway, leaving those mappings pointing left of the text they
describe, which failed the `sourceChar === targetChar` invariant in `Lines#getSourceMap`.

It now derives the shift by comparing the indentation the old and new `Lines` actually
render, so each mapping moves as far as its own line moved. `skipFirstLine` and
`noNegativeColumns` fall out of that comparison, making `Mapping#indent` 27 lines shorter.

Added a regression test that throws `Invariant failed` without the fix.

### DIFF
--- a/lib/lines.ts
+++ b/lib/lines.ts
@@ -194,9 +194,9 @@ export class Lines {
     if (this.mappings.length > 0) {
       const newMappings = lines.mappings;
       invariant(newMappings.length === 0);
-      this.mappings.forEach(function (mapping) {
-        newMappings.push(mapping.indent(width, skipFirstLine, true));
-      });
+      this.mappings.forEach(function (this: Lines, mapping) {
+        newMappings.push(mapping.indent(this, lines));
+      }, this);
     }
 
     return lines;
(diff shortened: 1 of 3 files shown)


eventualbuddha (approved):

eventualbuddha:
Thanks! Published in v0.23.15.

PR #1435, opened 2026-07-26
fix: preserve modern TypeScript syntax when printing

TheAlexLichter:
This PR ensures that modern TS modifiers are preserved during printing. These include (non-exhaustive):

- optional generic calls and typed binding patterns
- type-only imports/exports and import attributes
- import-type options, mapped modifiers, and type-query generics

Tests are added to cover the changes.


eventualbuddha (approved):

eventualbuddha:
Thanks! I’ll publish a new version soon.

eventualbuddha:
Published in v0.23.13.

PR #1436, opened 2026-07-26
feat: add oxc parser support

### LINKED ISSUE #1424: Support `oxc-parser`
Being [the fastest JS parser](https://github.com/oxc-project/bench-javascript-parser-written-in-rust) out there, it would be nice to support [oxc-parser](https://oxc.rs/docs/guide/usage/parser.html) in recast. The resulting AST is fully estree-compatible.

The only possible drawback is that the `loc` field is not available (at the time of writing, tracked in https://github.com/oxc-project/oxc/issues/10307).

TheAlexLichter:
Based on #1435 (optionally #1434)
Closes #1424 

This PR adds support for Oxc as parser. It normalizes Oxc AST nodes into Recast-compatible shapes.
To support currently unsupported modern fields, the PR also extends `ast-types` (can be upstreamed as well)
  
I've added various test types to make sure that the Oxc integration is working as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,7 +168,16 @@ const tsAst = recast.parse(source, {
 });
 ```
 
-**Note:** Some of these parsers import npm packages that Recast does not directly depend upon, so please be aware you may have to run `npm install @babel/parser` to use the `typescript`, `flow`, or `babel` parsers, or `npm install acorn` to use the `acorn` parser. Only Esprima is installed by default when Recast is installed.
+Oxc can be used for JavaScript, JSX, TypeScript, and TSX after installing
+`oxc-parser`:
+
+```js
+const oxcAst = recast.parse(source, {
+  parser: require("recast/parsers/oxc")
+});
+```
+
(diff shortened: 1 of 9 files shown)


eventualbuddha:
Thanks, @TheAlexLichter. I'm open to merging this, and overall I think the code looks pretty good. I don't love the idea of mutating the in-memory `ast-types`, though I get why you're doing it. I think it'd be best if we could get the updates into `ast-types` itself or just inline `ast-types` completely into `recast`.

@benjamn, do you have a preference here? I don't have write access or publish access to `ast-types`. Granting me access might be a short term fix, but longer-term maintenance of the project(s) should likely go to someone more invested in it.

TheAlexLichter:
If @benjamn is down for it, happy to either update `ast-types` 👍🏻
Was thinking about that too as another iteration.

TheAlexLichter:
@eventualbuddha do you have any preference on how to proceed?

benjamn:
@TheAlexLichter @eventualbuddha Take a look at https://github.com/benjamn/ast-types/releases/tag/v0.16.3 and let me know if I missed anything!

TheAlexLichter:
LGTM! ast-types 0.16.3 covers what the Oxc integration needs. I've updated the PR to require `ast-types@^0.16.3` and removed `_oxc_ast_types.ts` plus the shim-specific test.

PR #1440, opened 2026-08-17
fix(printer): preserve significant JSX whitespace around expressions when reprinting elements

ejhammond:
## Problem

When recast reprints a `JSXElement` subtree generically (for example because an identifier inside it changed), a significant space between text and an adjacent `{expression}` is dropped.

```js
const recast = require("recast");

const src = `function C() {
  return (
    <Text>hello {name} world</Text>
  );
}`;

const ast = recast.parse(src, { parser: require("recast/parsers/babel") });

// rename <Text> -> <Label>
recast.visit(ast, {
  visitJSXIdentifier(path) {
    if (path.node.name === "Text") path.node.name = "Label";
    this.traverse(path);
  },
});

console.log(recast.print(ast).code);
// Actual:   return (<Label>hello {name}world</Label>);
// Expected: return (<Label>hello {name} world</Label>);
```

The space before `world` disappears, which changes the rendered output (`hello world` becomes `helloworld`).

## Root cause

In the `JSXElement`/`JSXFragment` generic printer (`lib/printer.ts`), string `JSXText` children have their leading whitespace stripped unconditionally:

```js
if (/\S/.test(child.value)) {
  return child.value.replace(/^\s+/g, "");
}
```

That strip is meant to remove *source indentation* (the newline + spaces that precede text written on its own line, since indentation is reapplied later via `indentTail`). But it also removes a **significant inline space** that sits between a preceding sibling — typically a `{expression}` container — and the text. In the example above the third child is the `JSXText` `" world"`, and its leading space is stripped to `"world"`.

Per JSX text semantics, inline whitespace with no newline collapses to a single space and is preserved; only whitespace containing a newline is treated as insignificant formatting.

## Fix

Make the leading-whitespace collapse newline-aware:

```js
return child.value.replace(/^\s+/, (whitespace) =>
  /\n/.test(whitespace) ? "" : " ",
);
```

- Leading whitespace **containing a newline** → stripped entirely (unchanged behavior; indentation is reapplied by `indentTail`).
- Leading whitespace with **no newline** → collapsed to a single significant space and preserved.

Trailing-whitespace behavior is unchanged (the original code only touched leading whitespace).

## Tests

Added a `should preserve significant whitespace around JSX expressions` block in `test/jsx.ts` covering:

- text before an expression (`hello {name}`)
- text after an expression (`{name} world`)
- text on both sides (`hello {name} world`)
- multiple adjacent expressions (`a {x} b {y} c`)
- the newline-indentation case still collapses (regression guard)
- a no-change control that must round-trip byte-for-byte

Full suite: `npm test` passes (783 passing, 1 pre-existing pending). The existing `should not remove trailing whitespaces` and `should not double parentheses in Babel` tests, which exercise this same branch, continue to pass unchanged.

### DIFF
--- a/lib/printer.ts
+++ b/lib/printer.ts
@@ -1338,7 +1338,17 @@ function genericPrintNoParens(path: any, options: any, print: any) {
             typeof child.value === "string"
           ) {
             if (/\S/.test(child.value)) {
-              return child.value.replace(/^\s+/g, "");
+              // Collapse leading whitespace on JSXText children. Leading
+              // whitespace that contains a newline is source indentation and
+              // is safe to strip entirely (indentation is reapplied via
+              // indentTail below). Leading whitespace with no newline is
+              // significant inline whitespace adjacent to a sibling (typically
+              // a {expression} container); per JSX text semantics it collapses
+              // to a single space and must be preserved, otherwise e.g.
+              // `foo {bar} baz` would reprint as `foo {bar}baz`.
+              return child.value.replace(/^\s+/, (whitespace: string) =>
+                /\n/.test(whitespace) ? "" : " ",
+              );
             } else if (/\n\s*\n/.test(child.value)) {
               return "\n\n";
             } else if (/\n/.test(child.value)) {
--- a/test/jsx.ts
+++ b/test/jsx.ts
@@ -125,6 +125,96 @@ it("should not double parentheses in Babel", function () {
   );
 });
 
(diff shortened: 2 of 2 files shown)


ejhammond:
@eventualbuddha not sure about the contrib process here. Lmk if you need anything from me in order to get this fix landed :)

eventualbuddha:
Thanks! I'll have this out in a release soon.

